PR #78, opened 2015-01-15
added support for spark command

KarandeepJohar:
Could you please review it and guide me with the changes that are required

### DIFF
--- a/qds_sdk/commands.py
+++ b/qds_sdk/commands.py
@@ -336,11 +336,11 @@ class SparkCommand(Command):
         elif options.program is not None:
             if options.commandline is not None:
                 raise ParseError("Both commandline and program cannot be specified together", cls.optparser.format_help())
-        if options.commandline is not None:
+        if options.commandline is not None or options.script_location is not None:
             if options.language is not None:
-                raise ParseError("Both commandline and language cannot be specified together", cls.optparser.form)
+                raise ParseError("Both commandline/script location and language cannot be specified together", cls.optparser.form)
         elif options.language not in allowedlanglist:
-            raise ParseError("Please use one of".join(allowedlanglist), cls.optparser.format_help())
+            raise ParseError("Please use one of ".join(allowedlanglist), cls.optparser.format_help())
         if options.macros is not None:
             options.macros = json.loads(options.macros)
         v = vars(options)
--- a/qds_sdk/commands.py
+++ b/qds_sdk/commands.py
@@ -254,6 +254,98 @@ class HiveCommand(Command):
         v["command_type"] = "HiveCommand"
         return v
 
+class SparkCommand(Command):
+	
+	#{"program":prog,"language":"python","command_type":"SparkCommand","label":"praveen-spark"} 
+    usage = ("sparkcmd <submit|run> [options]")
+    allowedlandlist = ["python", "scala"]
+    optparser = GentleOptionParser(usage=usage)
+    optparser.add_option("--program", dest="program", help="program code")
+    
+    optparser.add_option("--cmdline", dest="commandline", help="command line for Spark")
+        
+    optparser.add_option("-f", "--script_location", dest="script_location",
+                         help="Path where spark program to run is stored. Can be S3 URI or local file path")
+
+    optparser.add_option("--macros", dest="macros",
+                         help="expressions to expand macros used in query")
+	
+    optparser.add_option("--tags", dest="tags",
+                         help="comma-separated list of tags to be associated with the query ( e.g., tag1 tag1,tag2 )")
+
+    optparser.add_option("--sample_size", dest="sample_size",
+                         help="size of sample in bytes on which to run query")
+
+    optparser.add_option("--cluster-label", dest="label", help="the label of the cluster to run the command on")
+	
+    optparser.add_option("--language", dest="language", help="the language used in the spark program, supported languages are scala and python")
+    
+    optparser.add_option("--label", dest="label_program",help="the label for the program")
+	
+    optparser.add_option("--notify", action="store_true", dest="can_notify", default=False, help="sends an email on command completion")
+
+    optparser.add_option("--name", dest="name", help="Assign a name to this query")
+           
+
+    @classmethod
+    def parse(cls, args):
+        """
+        Parse command line arguments to construct a dictionary of command
+        parameters that can be used to create a command
+
+        Args:
+            `args`: sequence of arguments
+
+        Returns:
+            Dictionary that can be used in create method
+
+        Raises:
+            ParseError: when the arguments are not correct
+        """
+        allowedlanglist = ["python", "scala"]
+        try:
+            (options, args) = cls.optparser.parse_args(args)
+            if options.program is None and options.script_location is None and options.commandline is None:
+                raise ParseError("One of script location or program or commandline must be specified", cls.optparser.format_help())
+        except OptionParsingError as e:
+            raise ParseError(e.msg, cls.optparser.format_help())
+        except OptionParsingExit as e:
+            return None
+
+        if options.script_location is not None:
+            if options.program is not None:
+                if options.commandline is not None:
+                    raise ParseError("All the three parameters program, script_location, and cmdline cannot be specified together", cls.optparser.format_help())
+                else:
+                    raise ParseError("Both Program and script_location cannot be specified", cls.optparser.format_help())
+            elif options.commandline is not None:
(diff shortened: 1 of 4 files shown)


qds_sdk/commands.py
```diff
@@ -254,6 +254,98 @@ def parse(cls, args):
...
+	
+    optparser.add_option("--tags", dest="tags",
+                         help="comma-separated list of tags to be associated with the query ( e.g., tag1 tag1,tag2 )")
+
+    optparser.add_option("--sample_size", dest="sample_size",
+                         help="size of sample in bytes on which to run query")
+
+    optparser.add_option("--cluster-label", dest="label", help="the label of the cluster to run the command on")
+	
+    optparser.add_option("--language", dest="language", help="the language used in the spark program, supported languages are scala and python")
```
mindprince: Can we use `choices` here?
https://docs.python.org/2/library/optparse.html#optparse.Option.choices

qds_sdk/commands.py
```diff
@@ -254,6 +254,98 @@ def parse(cls, args):
         v["command_type"] = "HiveCommand"
         return v
 
+class SparkCommand(Command):
+	
+	#{"program":prog,"language":"python","command_type":"SparkCommand","label":"praveen-spark"} 
+    usage = ("sparkcmd <submit|run> [options]")
+    allowedlandlist = ["python", "scala"]
+    optparser = GentleOptionParser(usage=usage)
+    optparser.add_option("--program", dest="program", help="program code")
```
mindprince: Needs a better help message.

mindprince: Also, I think we should ask for a program file here instead of program code.
No one is going to type all the program code in the command line.
There is an example of this type of interaction - search for `custom_config_file`.

mindprince:
@PraveenSeluka - please review.

qds_sdk/commands.py
KarandeepJohar: You can attach a program file using --script-location argument I think.

qubole: --script-location would be a location on s3 and not local box. For now we
can follow whatever other commands support via SDK.

qds_sdk/commands.py
```diff
@@ -254,6 +254,98 @@ def parse(cls, args):
         v["command_type"] = "HiveCommand"
         return v
 
+class SparkCommand(Command):
+	
+	#{"program":prog,"language":"python","command_type":"SparkCommand","label":"praveen-spark"} 
+    usage = ("sparkcmd <submit|run> [options]")
+    allowedlandlist = ["python", "scala"]
```
bharathbhushan1: Remove this line? Seems unused.

qds_sdk/commands.py
```diff
@@ -254,6 +254,98 @@ def parse(cls, args):
...
+        Returns:
+            Dictionary that can be used in create method
+
+        Raises:
+            ParseError: when the arguments are not correct
+        """
+        allowedlanglist = ["python", "scala"]
+        try:
+            (options, args) = cls.optparser.parse_args(args)
+            if options.program is None and options.script_location is None and options.commandline is None:
```
bharathbhushan1: Move this if() outside the try.

qds_sdk/commands.py
```diff
@@ -254,6 +254,98 @@ def parse(cls, args):
...
+                raise ParseError("One of script location or program or commandline must be specified", cls.optparser.format_help())
+        except OptionParsingError as e:
+            raise ParseError(e.msg, cls.optparser.format_help())
+        except OptionParsingExit as e:
+            return None
+
+        if options.script_location is not None:
```
bharathbhushan1: May be simpler to do something like below. I think such fine grained errors are probably not needed.
if (script_location xor program xor cmdline) {
  error(only one of script_location, program, cmdline can be specified.
}

qds_sdk/commands.py
```diff
@@ -254,6 +254,98 @@ def parse(cls, args):
...
+                                     cls.optparser.format_help())
+                options.script_location = None
+                options.query = q
+        elif options.program is not None:
+            if options.commandline is not None:
+                raise ParseError("Both commandline and program cannot be specified together", cls.optparser.format_help())
+        if options.commandline is not None or options.script_location is not None:
+            if options.language is not None:
+                raise ParseError("Both commandline/script location and language cannot be specified together", cls.optparser.form)
+        elif options.language not in allowedlanglist:
```
bharathbhushan1: When you implement my earlier review comment, this will probably be:
if (program and (lang is null or lang not in allowedlanglist) ERROR }

qds_sdk/commands.py
KarandeepJohar: Xor does make it simpler however, it doesn't cover the case when all three are given, which can be handled separately.

qds_sdk/commands.py
KarandeepJohar: So the "lang" is meaningless for other options i.e. script_location and cmdline?

KarandeepJohar: To frame it in a better way. When is the language options required?
For which of the the three input options is it necessary or optional?

qds_sdk/commands.py
```diff
@@ -257,6 +257,144 @@ def parse(cls, args):
...
+    allowedlanglist = ["python", "scala"]
+    
+    optparser = GentleOptionParser(usage=usage)
+    optparser.add_option("--program", dest="program", help="program code")
+    
+    optparser.add_option("--cmdline", dest="cmdline", help="command line for Spark")
+        
+    optparser.add_option("-f", "--script_location", dest="script_location",
+                         help="Path where spark program to run is stored. Can be S3 URI or local file path")
+
```
PraveenSeluka: lets remove supporting s3 uri - throw error

tests/test_command.py
```diff
@@ -290,6 +305,236 @@ def test_submit_sample_size(self):
...
+                 'cmdline': None,
+                 'command_type': 'SparkCommand',
+                 'arguments': None,
+                 'user_program_arguments': None,
+                 'can_notify': False,
+                 'script_location': None})
+    
+    def test_submit_user_program_arguments(self):
+        sys.argv = ['qds.py', 'sparkcmd', 'submit', '--language','scala','--program',
+                    "object HelloWorld {\n\n    def main(args: Array[String]) {\n        \n        println(\"Hello, \" + args(0))\n    \n    }\n}\n",
```
mindprince: Discussed with @PraveenSeluka,
There is no point of supporting this `--program`. No one is ever going to type the complete program while invoking qds.py.

So here is what we should do:
Instead of `--program`, `--cmdline` and `--script-location`.
We should only have `--cmdline` and `--program-location`. Because we can infer the language from the program location we can get rid of the `--language` option also. This will simplify the code a lot. And will also provide a better user experience.

qds_sdk/commands.py
```diff
@@ -257,6 +257,144 @@ def parse(cls, args):
...
+                         help="expressions to expand macros used in query")
+	
+    optparser.add_option("--tags", dest="tags",
+                         help="comma-separated list of tags to be associated with the query ( e.g., tag1 tag1,tag2 )")
+
+    optparser.add_option("--cluster-label", dest="label", help="the label of the cluster to run the command on")
+	
+    optparser.add_option("--language", dest="language", choices = allowedlanglist, help="the language used in the spark program, supported languages are scala and python")
+    
+    optparser.add_option("--label", dest="label_program",help="the label for the program")
```
mindprince: What is `label_program`?

tests/test_command.py
```diff
@@ -290,6 +305,236 @@ def test_submit_sample_size(self):
...
+                 'tags': None,
+                 'name': None,
+                 'program': "println(\"hello, world!\")",
+                 'cmdline': None,
+                 'command_type': 'SparkCommand',
+                 'arguments': None,
+                 'user_program_arguments': None,
+                 'can_notify': False,
+                 'script_location': None})
+
```
mindprince: There is no tests for the case where `script_location` is a local file. There should also be test cases to test whether we are determining the language correctly or not.

mindprince:
This also adds too many trailing white spaces.

qds_sdk/util.py
```diff
@@ -1,7 +1,6 @@
 import re
 import optparse
-
-
+from optparse import SUPPRESS_HELP as SUPPRESS_HELP
```
mindprince: Import it directly in commands.py?

mindprince:
Squash merged: https://github.com/qubole/qds-sdk-py/commit/1b5ecb83062bc41167b5a280de0d4d6639fc3f75

PR #80, opened 2015-01-28
fix urllib3 import error

hiyer:
Fix for import error when running qds.py.

### DIFF
--- a/qds_sdk/connection.py
+++ b/qds_sdk/connection.py
@@ -6,7 +6,7 @@ import json
 from requests.adapters import HTTPAdapter
 try:
     from requests.packages.urllib3.poolmanager import PoolManager
-except ImportError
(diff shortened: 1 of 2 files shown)


jsensarma:
should we specify a urllib3 dependency in setup.py?

when i had tried out this change in my vbox - it barfed on missing urllib3.

hiyer:
Yes, I was wondering if we need to specify the dependency somewhere. I will add this - any idea what would be a reasonable version to use (my laptop currently has 1.8.3, but PoolManager has been around since v1.0 of urllib3)?

jsensarma:
yeah - something very low so that we don't start failing unexpectedly. v1.0 sounds fine

mindprince:
Squash merged.

PR #84, opened 2015-02-22
optionally fetch the results when they are greater than 50MB

vogxn:
in certain cases the result_location is used to further create external
tables. but since the sdk returns raw results this is not doable. for
these i want to optionally fetch the result_location only and not read
the results from the s3 bucket and return to client.

### DIFF
--- a/qds_sdk/commands.py
+++ b/qds_sdk/commands.py
@@ -154,12 +154,13 @@ class Command(Resource):
         return r.text
 
 
-    def get_results(self, fp=sys.stdout, inline=True, delim=None):
+    def get_results(self, fp=sys.stdout, inline=True, delim=None, fetch=True):
         """
         Fetches the result for the command represented by this object
 
         Args:
             `fp`: a file object to write the results to directly
+            `fetch`: fetches the result if it is small, else returns the s3 bucket location
         """
         result_path = self.meta_data['results_resource']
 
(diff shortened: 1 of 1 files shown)


qds_sdk/commands.py
```diff
@@ -179,16 +180,21 @@ def get_results(self, fp=sys.stdout, inline=True, delim=None):
...
+                                            aws_secret_access_key=acc.storage_secret_key)
+
+                log.info("Starting download from result locations: [%s]" % ",".join(r['result_location']))
+                #fetch latest value of num_result_dir
+                num_result_dir = Command.find(self.id).num_result_dir
+                for s3_path in r['result_location']:
+                    # In Python 3, in this case, `fp` should always be binary mode.
+                    _download_to_local(boto_conn, s3_path, fp, num_result_dir, delim=delim)
+            else:
+                log.info("Returning Result Locations: [%s]" % ",".join(r['result_location']))
```
mindprince: The log line looks redundant.

qds_sdk/commands.py
```diff
@@ -154,12 +154,13 @@ def get_jobs_id(cls, id):
...
 
 
-    def get_results(self, fp=sys.stdout, inline=True, delim=None):
+    def get_results(self, fp=sys.stdout, inline=True, delim=None, fetch=True):
         """
         Fetches the result for the command represented by this object
 
         Args:
             `fp`: a file object to write the results to directly
+            `fetch`: fetches the result if it is small, else returns the s3 bucket location
```
mindprince: This statement needs improvement.
The results are always fetched when they are small. When the results are big, they would be fetched if this is true. Otherwise the s3 location would be returned.

qds_sdk/commands.py
vogxn: Will remove the log and write to the file directly.

As an aside - I think the get_results() behaviour of reading s3 blobs is not desirable. let me know what you think. This potentially starts reading large result sets (think >1GB) and starts downloading them. It doesn't have retries to handle AWS API failures either. Do you think fetch=False should be made default? Is this behaviour expected right now by users of the SDK?

mindprince: Yes. As we saw today. :-)

mindprince:
Squash merged: https://github.com/qubole/qds-sdk-py/commit/8261edac9ac397527b42753e78cade050e4ed022

PR #87, opened 2015-02-28
add support for cluster clone api

hiyer:


qds_sdk/cluster.py
```diff
@@ -127,15 +127,21 @@ def _parse_create_update(cls, args, action):
         argparser = ArgumentParser(prog="cluster %s" % action)
 
         create_required = False
+        label_required = False
+
```
mindprince: There is trailing whitespace here.

PR #91, opened 2015-03-17
chg: dev: Add support for cluster_manage_command

elricL:


tests/test_cluster.py
```diff
@@ -1478,5 +1478,27 @@ def test_minimal(self):
                                                     }
                                                 })
 
+class TestClusterManageCommands(QdsCliTestCase):
```
mindprince: Can you add some more negative tests:
- fail if both id and label are provided.
- fail if private_dns is provided with add_node.
- fail if private_dns is not provided with remove_node/update_node

qds_sdk/cluster.py
```diff
@@ -362,6 +362,33 @@ def clone(cls, cluster_id_label, cluster_info):
...
+                           help="execute on cluster with this label")
+
+      argparser.add_argument("--parameters",
+                 help="optional parameters to the manage command")
+
+      argparser.add_argument("--private_dns",
+                 help="the private_dns of the machine to be updated/removed")
+
+      argparser.add_argument("--command",
+                 help="the update command to be executed")
```
mindprince: Just for my understanding - what can be the various values for this argument?

elricL: Only one currently supported is replace. 
(More to be supported later such as, reboot etc.)

qds_sdk/cluster.py
```diff
@@ -362,6 +362,33 @@ def clone(cls, cluster_id_label, cluster_info):
...
+
+      group.add_argument("--id", dest="cluster_id",
+                           help="execute on cluster with this id")
+
+      group.add_argument("--label", dest="label",
+                           help="execute on cluster with this label")
+
+      argparser.add_argument("--parameters",
+                 help="optional parameters to the manage command")
+
```
vrajat: I dont think parameters are being used by any of the commands. If thats true, lets remove it from here.

qds_sdk/cluster.py
vrajat: You can use *choices* to restrict the list. Check docs for argparser.

qds_sdk/cluster.py
```diff
@@ -362,6 +362,33 @@ def clone(cls, cluster_id_label, cluster_info):
...
+      argparser.add_argument("--private_dns",
+                 help="the private_dns of the machine to be updated/removed")
```
vrajat: This is valid for replace only ? If yes, then put it under an if condition. Maybe it time to have a class hierarchy on the side. *ClusterApiCmdLine* Too much repetition. For e.g. check snapshot/restore_point. Anyway, cleanup for another day.

vrajat:
Please reopen PR vs *unreleased* branch. Maybe after all the current review comments addressed so that we dont have to repeat them in the new one.

qds_sdk/cluster.py
mindprince: I have some other changes in mind too: https://github.com/mindprince/qds-sdk-py/commit/65a667b9b8592079c5a1db8878b2e1e1d4b3eaf7
Now that more people are contributing, maybe we can do this.

vrajat: Wow! Thats a big deal! +1 on the intent

qds_sdk/cluster.py
elricL: Not currently. But hadoop/spark add might use those.

vrajat: I suggest you remove. There is a good chance we will have some way class hierarchy to set it up. Check @mindprince 's comment

mindprince:
Superseded by #98.

Issue #99, opened 2015-04-17
unreleased versions are not installed when setup has 1.4.3 as stable


vogxn:
The `setup.py`'s setup() method indicates the stable version to be installed. Because the unreleased branch of qds-sdk has `1.4.3` as the stable version, downstream tools that want the unreleased versions by specifying a github dependency pointed to the unreleased branch will fail to install. This is because setuptools finds `1.4.3` to be stable over `unreleased` and always override with what is found in PyPi. In order to  install `unreleased` versions the "stable" version in the unreleased branch should be made `unreleased`

PR #100, opened 2015-04-17
Fix for #99 : Unreleased versions

vogxn:
The `setup.py`'s setup() method indicates the stable version to be installed. Because the unreleased branch of qds-sdk has `1.4.3` as the stable version, downstream tools that want the unreleased versions by specifying a github dependency pointed to the unreleased branch will fail to install. This is because setuptools finds `1.4.3` to be stable over `unreleased` and always override with what is found in PyPi. In order to  install `unreleased` versions the "stable" version in the unreleased branch should be made `unreleased`


vogxn:
Need to open against unreleased.

PR #102, opened 2015-04-20
Add connect and read timeouts for http requests

mindprince:
Also add Timeout Exception to the list of exceptions for which GET requests are retried.


rjainqb:
LGTM

PR #103, opened 2015-04-24
new: usr: support for creating spark clusters

vogxn:
this commit adds support for creation of spark clusters from the qds-sdk

### DIFF
--- a/qds_sdk/cluster.py
+++ b/qds_sdk/cluster.py
@@ -215,9 +215,7 @@ class Cluster(Resource):
                              action="store_false",
                              default=None,
                              help="Use hadoop1 instead of hadoop2. This is the default.")
-
-        spark = hadoop_group.add_mutually_exclusive_group()
-        spark.add_argument("--use-spark",
+        hadoop2.add_argument("--use-spark",
                            dest="use_spark",
                            action="store_true",
                            default=None,
--- a/qds_sdk/cluster.py
+++ b/qds_sdk/cluster.py
@@ -216,6 +216,13 @@ class Cluster(Resource):
                              default=None,
                              help="Use hadoop1 instead of hadoop2. This is the default.")
 
+        spark = hadoop_group.add_mutually_exclusive_group()
+        spark.add_argument("--use-spark",
+                           dest="use_spark",
+                           action="store_true",
+                           default=None,
+                           help="Turn on spark for this cluster")
+
         spot_group = argparser.add_argument_group("spot instance settings" +
                     " (valid only when slave-request-type is hybrid or spot)")
         spot_group.add_argument("--maximum-bid-price-percentage",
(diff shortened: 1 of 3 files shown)


qds_sdk/cluster.py
```diff
@@ -216,6 +216,13 @@ def _parse_create_update(cls, args, action):
                              default=None,
                              help="Use hadoop1 instead of hadoop2. This is the default.")
 
+        spark = hadoop_group.add_mutually_exclusive_group()
+        spark.add_argument("--use-spark",
+                           dest="use_spark",
+                           action="store_true",
+                           default=None,
+                           help="Turn on spark for this cluster")
```
mindprince: Add a parameter which let's me edit a spark cluster to become a non spark cluster.

mindprince:
Needs tests.

qds_sdk/cluster.py
vrajat: I have my doubts. This is required for testing. However in production, what does it mean to make it a non-spark cluster or non-hbase cluster ? The affects can be devastating. Should we make it difficult to change type of clusters ? 
Users can always create new clusters and re-label them properly.

mindprince: Valid point.

Though saying `--dont-use-spark` would show their intention pretty clearly. I mean we can do that from the UI by changing the cluster type - so wondering whether the SDK should be crippled in this regard.

I don't have any strong opinions on this other than parity between UI and SDK. So, your decision.

vrajat: I am going to start a discussion on crippling the UI too. That was written
before the current surge in types of clusters. We spoke about designing
something proper in a PR RG had started.

On Wed, Apr 29, 2015 at 1:05 AM, Rohit Agarwal <notifications@github.com>
wrote:

> In qds_sdk/cluster.py
> <https://github.com/qubole/qds-sdk-py/pull/103#discussion_r29279913>:
>
> > @@ -216,6 +216,13 @@ def _parse_create_update(cls, args, action):
> >                               default=None,
> >                               help="Use hadoop1 instead of hadoop2. This is the default.")
> >
> > +        spark = hadoop_group.add_mutually_exclusive_group()
> > +        spark.add_argument("--use-spark",
> > +                           dest="use_spark",
> > +                           action="store_true",
> > +                           default=None,
> > +                           help="Turn on spark for this cluster")
>
> Valid point.
>
> Though saying --dont-use-spark would show their intention pretty clearly.
> I mean we can do that from the UI by changing the cluster type - so
> wondering whether the SDK should be crippled in this regard.
>
> I don't have any strong opinions on this other than parity between UI and
> SDK. So, your decision.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/qubole/qds-sdk-py/pull/103/files#r29279913>.
>



-- 
Rajat Venkatesh | Engg Lead
Qubole Inc | www.qubole.com

mindprince:
Looks good now.

mindprince:
Squash merged: https://github.com/qubole/qds-sdk-py/commit/2460284e5eafb9071d8f6cbd4a2d882ee1e32cac

mindprince:
@PraveenSeluka I think this can also be merged in the master branch?

PR #109, opened 2015-06-02
Adds SQL support in Spark Command

sourabh912:
Jira: https://qubole.atlassian.net/browse/SDK-8

@mindprince @PraveenSeluka

### DIFF
--- a/qds_sdk/commands.py
+++ b/qds_sdk/commands.py
@@ -295,12 +295,14 @@ class HiveCommand(Command):
 class SparkCommand(Command):
 
     usage = ("sparkcmd <submit|run> [options]")
-    allowedlanglist = ["python", "scala"]
+    allowedlanglist = ["python", "scala", "sql"]
 
     optparser = GentleOptionParser(usage=usage)
     optparser.add_option("--program", dest="program",help=SUPPRESS_HELP)
 
     optparser.add_option("--cmdline", dest="cmdline", help="command line for Spark")
+
+    optparser.add_option("--sql", dest="sql", help="sql for Spark")
 
     optparser.add_option("-f", "--script_location", dest="script_location",
                          help="Path where spark program to run is stored. Has to be a local file path")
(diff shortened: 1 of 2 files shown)


qds_sdk/commands.py
```diff
@@ -295,13 +295,15 @@ def parse(cls, args):
 class SparkCommand(Command):
 
     usage = ("sparkcmd <submit|run> [options]")
-    allowedlanglist = ["python", "scala"]
+    allowedlanglist = ["python", "scala", "sql"]
```
mindprince: Is this needed?

qds_sdk/commands.py
```diff
@@ -341,25 +343,38 @@ def validate_program(cls, options):
...
+        bool_sql = options.sql is not None
+        bool_other_options = options.script_location is not None or options.program is not None or options.cmdline is not None
+
+        # if both are false then no option is specified => raise PraseError
+        # if both are true then atleast two option specified => raise ParseError
+        if bool_sql == bool_other_options:
+            raise ParseError("Exactly One of script location or program or cmdline or sql should be specified", cls.optparser.format_help())
+        if bool_sql:
+            if options.language is None:
+                raise ParseError("Unspecified language for Sql query", cls.optparser.format_help())
```
mindprince: You don't need to send `language: sql`. So this error should be the other way around.

tests/test_command.py
```diff
@@ -407,6 +410,32 @@ def test_submit_script_location_local_java(self):
...
+                     'program': None,
+                     'cmdline':None,
+                     'command_type': 'SparkCommand',
+                     'arguments': None,
+                     'user_program_arguments': None,
+                     'can_notify': False,
+                     'script_location': None})
+
+    def test_submit_sql_no_program(self):
+        sys.argv = ['qds.py', 'sparkcmd', 'submit', '--sql', 'show dummy']
```
mindprince: This should be the positive test and the other one should raise an error.

qds_sdk/commands.py
sourabh912: I have tried to send all the parameters which we are sending from the Analyze UI and language is one of them. I will remove 'language' and run the query again and see if it succeeds.

PR #113, opened 2015-07-22
Refresh table added to SDK

sameerkedia:


mindprince:
This is incomplete. Fails existing tests. Doesn't add any new tests. The help messages are random strings. And the purpose of the PR is also not clear. Please fix.

sameerkedia:
Edited the help messages. Added a test for refresh table. Please see if it's correct.

swatisinghi:
@mindprince, Sameer had changed the existing hive command option through the SDK to handle refresh table where as on the QDS UI Refresh table shows up as a separate command type in the drop down (although it is handled as a Hive command in the rails layer). So should we keep the behavior consistent with the UI and expose another refresh table command type through the SDK too (it can internally fire a hive command with the correct params). 

Please Suggest.

harshshah87:
This looks like an old abandoned PR. If someone is working on it or ready to work on it, reopen it.

chattarajoy:
This looks like an old abandoned PR. If someone is working on it or ready to work on it, reopen it.

PR #112, opened 2015-07-22
Support for data sharing through sdk

jainavi:
Support for data sharing through sdk

PR #116, opened 2015-07-27
Support for data sharing through sdk

jainavi:
Support for data sharing through sdk


mindprince:
Closing - same as #112

PR #117, opened 2015-08-26
Account creation SDK

jainavi:


mindprince:
This PR is wrong just like #112

PR #120, opened 2015-09-07
SDK-25: QDS SDK changes to support cluster API 1.3

prakharjain09:


bin/qds.py
```diff
@@ -183,107 +183,129 @@ def checkargs_cluster_id_label(args):
...
-                             str(e))
-            usage()
-    cluster_info.set_presto_settings(arguments.enable_presto,
-                                     presto_custom_config)
+        cluster_info.set_presto_settings(arguments.enable_presto,
+                                         presto_custom_config)
 
     return cluster_info
 
+def _read_file(file_path, file_name):
```
mindprince: This is nice.

qds_sdk/cluster.py
```diff
@@ -286,7 +330,9 @@ def _parse_create_update(cls, args, action):
                                  default=None,
                                  help="don't encrypt the ephemeral drives on" +
                                       " the instance",)
-        security_group.add_argument("--customer-ssh-key",
+
+        customer_ssh_key_parameter = "--ssh-public-key" if (api_version >= 1.3) else "--customer-ssh-key"
```
mindprince: Why is this different based on api version? We can keep the command line flag same even if the API argument is different.

bin/qds.py
```diff
@@ -183,107 +183,129 @@ def checkargs_cluster_id_label(args):
...
+    custom_config = _read_file(arguments.custom_config_file, "custom config file")
+    presto_custom_config = _read_file(arguments.presto_custom_config_file, "presto custom config file")
+    fairscheduler_config_xml = _read_file(arguments.fairscheduler_config_xml_file, "config xml file")
+    ssh_public_key = _read_file(arguments.customer_ssh_key_file, "customer ssh key file")
+
+    cluster_info = None
+    if api_version >= 1.3:
+        cluster_info = ClusterInfoV13(arguments.label, api_version)
+        cluster_info.set_cluster_info(aws_access_key_id=arguments.aws_access_key_id,
+                         aws_secret_access_key=arguments.aws_secret_access_key,
```
mindprince: This is indented wrong.

Indent with twice the tab size or line up with the argument in the last line. https://www.python.org/dev/peps/pep-0008/#indentation

qds_sdk/cluster.py
```diff
@@ -174,35 +174,79 @@ def _parse_create_update(cls, args, action):
...
                                   help="location of file containg custom" +
                                        " hadoop configuration overrides")
-        hadoop_group.add_argument("--slave-request-type",
-                                  dest="slave_request_type",
-                                  choices=["ondemand", "spot", "hybrid"],
-                                  help="purchasing option for slave instaces",)
         hadoop_group.add_argument("--use-hbase", dest="use_hbase",
                                   action="store_true", default=None,
                                   help="Use hbase on this cluster",)
+        if api_version >= 1.3:
```
mindprince: Can we move this to the end after all the common arguments are done.

mindprince: Oh looks like these are for hadoop_group/node_config_group and that's why they are here. It's fine then.

tests/test_cluster.py
```diff
@@ -931,6 +931,453 @@ def test_fairscheduler_default_pool(self):
                     }
                 })
 
+    def test_ebs_volume_type_v13(self):
+        sys.argv = ['qds.py', '--version', 'v13','cluster', 'create', '--label', 'test_label',
```
mindprince: Shouldn't this be v1.3 instead of v13?

mindprince: Same for all the tests.

qds_sdk/cluster.py
prakharjain09: We have changed the command line flag in the new api version from "customer_ssh_key" to "ssh_public_key". I think its better to change this flag also when someone is using >= 1.3.

bin/qds.py
prakharjain09: updated.

qds_sdk/cluster.py
mindprince: Yeah, I know that the API expects the parameter to be named `ssh_public_key` now. But I think the command line flag should still be called `--customer-ssh-key`.

The reason for this is backwards compatibility. At some point in the future, we would change the default version from `v1.2` to `v1.3`.

If we change this command-line flag, all the command lines of the following form would need to be updated:
```
qds.py cluster create --label test_label
                      --access-key-id aki
                      --secret-access-key sak
                      --customer-ssh-key file
```
However, if we keep calling the command-line flag `--customer-ssh-key`, we would maintain command-line backwards compatibility.

I would also suggest that you go through this PR and look for other cases which may have introduced incompatibilities like this and fix them.

prakharjain09: Updated this. Also looked at the PR. Didnt find any other case.

mindprince:
Squash merged: https://github.com/qubole/qds-sdk-py/commit/7a3b8ec5925df2d4238cb5d5c715c610369d7c9d

PR #121, opened 2015-09-17
SDK-31: Ensure all our SDKs comply with backoff requirements in our new SLA

prakharjain09:
our SLA document will now have the following clause:

"Back-off Requirements" means, when an error occurs, the Application is responsible
waiting for a period of time before issuing another request. This means that after the first 
error, there is a minimum back-off interval of 1 second and for each consecutive error, the
back-off interval increases exponentially (by a factor of 2) up to 32 seconds.

we have to make sure that all our SDKs are in compliance. The errors in question here are all 5xx error codes.

### DIFF
--- a/qds_sdk/connection.py
+++ b/qds_sdk/connection.py
@@ -43,20 +43,23 @@ class Connection:
             self.session = requests.Session()
             self.session.mount('https://', MyAdapter())
 
-    @retry((RetryWithDelay, requests.Timeout), tries=6, delay=30, backoff=2)
+    @retry((RetryWithDelay, requests.Timeout), tries=12, delay=1, backoff=2, max_delay=32)
     def get_raw(self, path, params=None):
         return self._api_call_raw("GET", path, params=params)
 
-    @retry((RetryWithDelay, requests.Timeout), tries=6, delay=30, backoff=2)
+    @retry((RetryWithDelay, requests.Timeout), tries=12, delay=1, backoff=2, max_delay=32)
     def get(self, path, params=None):
         return self._api_call("GET", path, params=params)
 
+    @retry((RetryWithDelay, requests.Timeout), tries=12, delay=1, backoff=2, max_delay=32)
     def put(self, path, data=None):
         return self._api_call("PUT", path, data)
 
+    @retry((RetryWithDelay, requests.Timeout), tries=12, delay=1, backoff=2, max_delay=32)
     def post(self, path, data=None):
         return self._api_call("POST", path, data)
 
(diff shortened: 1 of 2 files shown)


qds_sdk/connection.py
```diff
@@ -43,20 +43,23 @@ def __init__(self, auth, base_url, skip_ssl_cert_check, reuse=True):
             self.session = requests.Session()
             self.session.mount('https://', MyAdapter())
 
-    @retry((RetryWithDelay, requests.Timeout), tries=6, delay=30, backoff=2)
```
prakharjain09: @mindprince 
1. currently we are raising RetryWithDelay only for 449, 502, 503, 504 error codes. should we do it for all 5xx codes?
2.  what should be the value of tries?

qds_sdk/connection.py
```diff
@@ -43,20 +43,23 @@ def __init__(self, auth, base_url, skip_ssl_cert_check, reuse=True):
...
+    @retry((RetryWithDelay, requests.Timeout), tries=12, delay=1, backoff=2, max_delay=32)
     def get(self, path, params=None):
         return self._api_call("GET", path, params=params)
 
+    @retry((RetryWithDelay, requests.Timeout), tries=12, delay=1, backoff=2, max_delay=32)
     def put(self, path, data=None):
         return self._api_call("PUT", path, data)
 
+    @retry((RetryWithDelay, requests.Timeout), tries=12, delay=1, backoff=2, max_delay=32)
```
mindprince: We can't retry on POST because it is not idempotent.
Not sure if we should retry on PUT/DELETE either. Will need to confirm with our rails team that they don't do anything which shouldn't be retried.

qds_sdk/connection.py
```diff
@@ -43,20 +43,23 @@ def __init__(self, auth, base_url, skip_ssl_cert_check, reuse=True):
             self.session = requests.Session()
             self.session.mount('https://', MyAdapter())
 
-    @retry((RetryWithDelay, requests.Timeout), tries=6, delay=30, backoff=2)
+    @retry((RetryWithDelay, requests.Timeout), tries=12, delay=1, backoff=2, max_delay=32)
     def get_raw(self, path, params=None):
         return self._api_call_raw("GET", path, params=params)
 
-    @retry((RetryWithDelay, requests.Timeout), tries=6, delay=30, backoff=2)
+    @retry((RetryWithDelay, requests.Timeout), tries=12, delay=1, backoff=2, max_delay=32)
```
mindprince: I don't like the new retry intervals. Let's discuss on the JIRA.

mindprince:
Closing this after discussion on the JIRA.

PR #122, opened 2015-09-18
SDK-21: Set a limit on minimum polling interval

prakharjain09:


qds_sdk/qubole.py
```diff
@@ -43,6 +46,9 @@ def configure(cls, api_token,
         cls._auth = QuboleAuth(api_token)
         cls.api_token = api_token
         cls.base_url = api_url.rstrip('/') + '/' + version
+        if poll_interval < Qubole.MIN_POLL_INTERVAL:
+            sys.stderr.write("Poll interval cannot be less than %s seconds.\n" % Qubole.MIN_POLL_INTERVAL)
+            sys.exit(2)
```
mindprince: Don't exit with error, if you do, it will break for people who are currently using a very low value.
Instead if the poll_interval is less than Qubole.MIN_POLL_INTERVAL, set the poll interval to MIN_POLL_INTERVAL and display a warning log message that we have done so.

In general, while making changes to the SDK, be very careful about not changing the behavior in ways that will break existing workflows.

qds_sdk/qubole.py
```diff
@@ -18,6 +19,8 @@ class Qubole:
     configuration parameters for QDS.
     """
 
+    MIN_POLL_INTERVAL = 3
```
prakharjain09: @mindprince I was using 3 sec for my testing. Forgot to add a comment on friday for the same. What minimum polling interval should we use? 5 sec?

mindprince:
Squash merged: https://github.com/qubole/qds-sdk-py/commit/adc3979d82fb828aa2de178f8e396b1c633f12e0

PR #128, opened 2015-10-27
SDK-55: Support for managing spark jobserver apps and submitting snippets to them.

mindprince:


mindprince:
cc - @PraveenSeluka @bharathbhushan1 @guptarajat

tests/test_command.py
```diff
@@ -680,6 +694,84 @@ def test_submit_R_program(self):
                  'can_notify': False,
                  'script_location': None})
 
+    def test_submit_program_to_app(self):
+        sys.argv = ['qds.py', 'sparkcmd', 'submit', '--language', 'scala',
+                    '--program', 'sc.version', '--app-id', '1']
```
PraveenSeluka: why do we have app-id as 1 here ? (Just because we will have this app id basically ?). Can we write a test which creates app, submits snippets and then deletes it.

PR #129, opened 2015-11-20
QBOL-4796 API support for role instance profile

yogesh2021:
Adding sdk support for a new cluster config field: `role_instance_profile`, will be available for all api versions. This field will be used only when IAM Roles are enabled. Also removing mandatory check for providing aws keys while creating cluster.

@guptavishal @mindprince @itsvikramagr @msumit


mindprince:
So, is this new field mutually exclusive with the AWS/Keys fields? Wondering if we can add a mutually exclusive check to make sure both are not present. Also can we check that one of these is definitely present.

yogesh2021:
Yes, this field is mutually exclusive of AWS Keys. The idea is to actually remove such checks from the qds-sdk and let tapp code handles it. In the latest commit, we have only removed the check from AWS Keys.

msumit:
lgtm

mindprince:
There will always be server side checks. But SDKs and CLIs are more helpful if they don't let the users commit the mistake at all. That's why these checks are important on the client side as well. So, if it's not too difficult and complicated to add that logic here, I would suggest we add it. If it makes the code exceedingly complicated, then we can skip.

yogesh2021:
We cannot ensure that the user doesn't commit mistake at all, as an example what if user gives "xyz" access key. The current implementation doesn't even check that. It will be bad to put business logics inside sdks such as like ours where they are meant to provide an interface to our backend.

mindprince:
We cannot ensure but we can do our best effort.

guptavishal:
lgtm

strong +1 for not duplicating the same logic at multiple places in this case.

mindprince:
I have squash merged this to the unreleased branch. https://github.com/qubole/qds-sdk-py/commit/c8775a9f789ae329098b7a6d4f7881ee533e1003

Let me know when this is available on production and I will merge it to the master branch and make a release.

PR #132, opened 2015-12-24
Add --print-logs-live for friendly running

dfrankow:
It's only friendly to show someone what's happening as a long command runs.

### DIFF
--- a/qds_sdk/commands.py
+++ b/qds_sdk/commands.py
@@ -87,9 +87,22 @@ class Command(Resource):
         Returns:
             Command object
         """
+        log_lines_consumed = 0
+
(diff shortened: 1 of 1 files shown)


dfrankow:
The tests don't pass, but something like this seems civilized. Please consider it.

mindprince:
Thanks @dfrankow
I agree, it is a very reasonable feature request.
I will confirm with the team if there is an API which will help us do it in a way better than downloading the complete logs again and again.

dfrankow:
Thanks!

harshshah87:
A PR has opened for live logging support.
https://github.com/qubole/qds-sdk-py/pull/134

Will close this one.

PR #134, opened 2015-12-30
Adding live log support

msumit:
dependent on https://bitbucket.org/qubole/qbol/pull-requests/2055/tapp-changes-to-provide-live-log


qds_sdk/commands.py
```diff
@@ -87,10 +87,26 @@ def run(cls, **kwargs):
         Returns:
             Command object
         """
+        err_pointer, tmp_pointer, new_bytes = 0, 0, 0
         cmd = cls.create(**kwargs)
         while not Command.is_done(cmd.status):
             time.sleep(Qubole.poll_interval)
             cmd = cls.find(cmd.id)
+            if kwargs.get('print_logs_live', False):
+                log, err_length, tmp_length = cmd.get_log_partial(err_pointer, tmp_pointer)
```
harshshah87: What is err_length and tmp_length here? And what will be the contents of logs? is it the complete log text or partial?

harshshah87:
The changes are going to be revisited for testing it extensively.

qds_sdk/commands.py
```diff
@@ -87,10 +87,26 @@ def run(cls, **kwargs):
...
+
+                if err_length != "0":
+                    err_pointer += int(err_length)
+                    new_bytes = int(err_length) + int(tmp_length) - new_bytes - tmp_pointer
```
yogesh2021: We don't need new_bytes in the calculation. tmp_pointer has already taken care of that in the previous iteration.

qds_sdk/commands.py
```diff
@@ -87,10 +87,26 @@ def run(cls, **kwargs):
...
+                    tmp_pointer = int(tmp_length)
+
+                else:
+                    tmp_pointer += int(tmp_length)
+                    new_bytes = int(tmp_length)
+
+                if len(log) > 0 and new_bytes > 0:
+                    data = list(log)
+                    print >>sys.stderr, "".join(data[-new_bytes:])
```
yogesh2021: I guess we don't need to do list operation on the log variable and then do join operation. we can simply do log[-new_bytes:].

qds_sdk/commands.py
```diff
@@ -87,10 +87,26 @@ def run(cls, **kwargs):
         Returns:
             Command object
         """
+        err_pointer, tmp_pointer, new_bytes = 0, 0, 0
         cmd = cls.create(**kwargs)
         while not Command.is_done(cmd.status):
```
yogesh2021: It could be possible that command is done but in the previous iteration we didn't get full remaining logs as it was still running. So it would be good to fetch the logs once again and apply the same logic as in while loop. We have seen such cases in UI.

adeshr: If command is completed, we cannot apply same logic because when a command gets completed, subsequent calls to logs-api returns complete logs instead of partial.

yogesh2021: @adeshr 
logs api work as told by client :). The reason it passes full logs is because we pass err_pointer as 0 and hence it passes entire log. This will happen here as well but we are handling this using logic inside:

> if err_length != "0":
new_bytes = int(err_length) + int(tmp_length) - new_bytes - tmp_pointer

Please tell if I am missing anything.

msumit: In UI we just delete the whole dom and re-populate it with new content. In cli I don't think its possible.

qds_sdk/commands.py
msumit: will change it.

qds_sdk/commands.py
msumit: may be.. also all these complex calculations are done for composite cmd, and afaik composite cmds can't be run from cli..

qds_sdk/commands.py
adeshr: @yogesh2021 if the command is completed, then err_length/tmp_length are ignored by tapp and complete logs are returned.

@msumit maybe we can add an `if` clause after `while` loop that checks if command is completed then compare the log length of the previous api-call with the current one. If there is any difference between log-length (err_length + tmp_length), then print that difference of logs(difference is to be printed from err_file)?

yogesh2021: I see. when command is finished api returns logs from cache / download from s3. In those cases err_length and tmp_length are 0 which are not correct indicators. Secondly is it safe to get all the logs once the command is finished, they could be very large. How are we handling this currently ? @msumit

qds_sdk/commands.py
yogesh2021: Yes, this is largely meant for composite command. This logic is helpful in more case, command might have finished when you called logs api, so it may return non-empty err logs and empty tmp logs.

PR #135, opened 2016-01-11
fix: usr: SDK-87: Fix code related to Nezha-Entities

adeshr:
Changed help messages wherever necessary, added a sub-parser for update code along with a few minor modifcations.

### DIFF
--- a/qds_sdk/nezha.py
+++ b/qds_sdk/nezha.py
@@ -15,7 +15,8 @@ class NezhaCmdLine(CmdLine):
                         'measures' : 'NezhaMeasure',
                         'partitions' : 'NezhaPartition',
                         'dimensions' : 'NezhaDimension',
-                        'data_sources' : 'NezhaDataSource'}
+                        'data_sources' : 'NezhaDataSource',
+                        'default_data_source' : 'NezhaDefaultDS'}
     classname = ""
 
     @staticmethod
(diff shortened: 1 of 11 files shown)


qds_sdk/nezha.py
```diff
@@ -6,169 +6,251 @@
...
-    """
+    qds_sdk.NezhaDimension: is the class for retrieving and
+    manipulating Qubole's NezhaDimension.
 
-    """ endpoint for data source is  /nezha_data_sources endpoint"""
+    Endpoint for dimensions is: /nezha_cubes/{nezha_cube_id}/nezha_dimensions.
+    User needs to set it before using.
+    """
 
     rest_entity_path = ""
```
mindprince: Why is this empty?

adeshr: NezhaDimension/NezhaMeasure were needed in PyApi test and rest_entity_path needs NezhaCube Id. So, before using it there, we are setting rest_entity_path based on cube-id.
No command-line support is given for Dimensions/Measures.

mindprince:
@vrajat Shouldn't this be called quark now?

qds_sdk/nezha.py
```diff
@@ -6,169 +6,251 @@
...
+        parser_partition.add_argument(
+            "--table_name", dest="table_name", help="Table name")
+        parser_partition.add_argument(
+            "--destination_id", dest="destination_id", help="Data Source id")
+        parser_partition.add_argument("--query", dest="query", help="Query")
+
+        parser_ds = update_subparsers.add_parser(
+            "default_datasource", help="update default-datasource")
+        parser_ds.add_argument(
+            "default_datasource_id", help="Numeric id of the DataSource")
```
mindprince: What does `update default_datasource ID` do? Does it change the default datasource to be a different ID? If that's the case, we should explain it in a better way.

PR #139, opened 2016-02-29
Added sdk for user (invite , enable ,disable)

tanishgupta1:


qds_sdk/user.py
```diff
@@ -0,0 +1,91 @@
...
+import argparse
+import json
+
+
+class UserCmdLine:
+    @staticmethod
+    def parsers():
+        argparser = argparse.ArgumentParser(
+            prog="qds.py user",
+            description="Account Creation for Qubole Data Service.")
```
sourabh912: I believe this class is for User (and not accounts) related operations

tanishgupta1: fixed

bin/qds.py
```diff
@@ -570,6 +577,9 @@ def main():
     if a0 == "nezha":
         return nezhamain(args)
 
+    if a0 == "user":
+        return usermain(args)
+
     cmdset = set(CommandClasses.keys())
     sys.stderr.write("First command must be one of <%s>\n" %
                      "|".join(cmdset.union(["cluster", "action", "scheduler", "report",
```
harshshah87: Add "user" here.

qds_sdk/user.py
```diff
@@ -0,0 +1,91 @@
...
+
+        # To invite a new user
+        invite = subparsers.add_parser(
+            "invite", help="Invite a new user")
+        invite.add_argument(
+            "--email", dest="invitee_email", required=True,
+            help="Email address of the new user to be added to the Qubole account")
+        invite.add_argument(
+            "--accountid", dest="account", required=True,
+            help="Account ID of the current user")
```
msumit: shouldn't we take the account_id from requesting xauth token itself?

tests/test_user.py
```diff
@@ -0,0 +1,92 @@
...
+
+    def test_invite_no_email(self):
+        sys.argv = ['qds.py', 'user', 'invite',
+                    '--accountid', '34',
+                    '--groups', "system-admin"]
+        print_command()
+        Connection._api_call = Mock(return_value={})
+        with self.assertRaises(SystemExit):
+            qds.main()
+
```
msumit: can u test a case when more than 1 group name is passed?

qds_sdk/user.py
```diff
@@ -0,0 +1,91 @@
...
+
+    rest_entity_path = "users"
+
+    @classmethod
+    def invite(cls, path, data):
+        conn = Qubole.agent()
+        return conn.post(cls.element_path(path), data)
+
+
+class Account(Resource):
```
msumit: why not add this class in `accout.py` file only?

qds_sdk/user.py
```diff
@@ -0,0 +1,91 @@
...
+        subparsers = argparser.add_subparsers(title="user operations")
+
+        # To invite a new user
+        invite = subparsers.add_parser(
+            "invite", help="Invite a new user")
+        invite.add_argument(
+            "--email", dest="invitee_email", required=True,
+            help="Email address of the new user to be added to the Qubole account")
+        invite.add_argument(
+            "--accountid", dest="account", required=True,
```
sourabh912: replace it with --account-id

qds_sdk/user.py
sourabh912: +1

qds_sdk/user.py
sourabh912: +1

tanishgupta1: rest_entity_path for Account in account.rb is "account" and in case of disable enable user it is "accounts" so declared separately here for now.

tests/test_user.py
tanishgupta1: done

PR #145, opened 2016-03-22
fixes #143 in qubole/qds-sdk-py

### LINKED ISSUE #143: ClusterInfo v1.3 is not very wieldy and extendible
As a part of this [commit](https://github.com/qubole/qds-sdk-py/commit/7a3b8ec5925df2d4238cb5d5c715c610369d7c9d) the `ClusterInfoV13` class was introduced to deal with the newer version of the cluster API. The previous form `ClusterInfo` had setters for each property (spot, security settings, ec2 settings etc) to update the specific parts of the cluster. Eg: If I needed to update only the spot instance properties I had only to call `set_spot_instance_settings(..)`. Now I have to call the long form `set_cluster_info(..)` on each update. I cannot call the double underscored `__set_spot_instance_settings(..)` anymore since these are changed into `_ClusterInfoV13__set_spot_instance_settings(..)` because of python's [name mangling](https://docs.python.org/2/tutorial/classes.html#private-variables-and-class-local-references). This is supremely frustrating and breaks a lot of test code that we have built up which needs to be rewritten as `__Cluster_set_spot_instance_settings(..)` or the long form update `Cluster.set_cluster_info(..)` for each update call to a cluster object. 

Is it possible to convert these setters to [`@property`-ies](http://stackoverflow.com/a/7456865) because that's how setters are exposed in Python. Was there a strong reason to hide the v1.3 setters deviating from the norm that was in v1.2 where setters were public?

vogxn:
double underscored names are mangled by python as they are seen as
private fields. this is not suitable when importing the sdk and writing
factories for various qubole resources as there is no easy way to update
specific parts of the cluster info object.

### DIFF
--- a/qds_sdk/cluster.py
+++ b/qds_sdk/cluster.py
@@ -1047,18 +1047,18 @@ class ClusterInfoV13():
         self.disallow_cluster_termination = disallow_cluster_termination
         self.enable_ganglia_monitoring = enable_ganglia_monitoring
         self.node_bootstrap_file = node_bootstrap_file
-        self.__set_node_configuration(master_instance_type, slave_instance_type, initial_nodes, max_nodes, slave_request_type, fallback_to_ondemand)
-        self.__set_ec2_settings(aws_access_key_id, aws_secret_access_key, aws_region, aws_availability_zone, vpc_id, subnet_id,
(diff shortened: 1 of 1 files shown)


vogxn:
Will be closing this PR and re-raising against the unreleased branch

PR #147, opened 2016-03-25
fixes #144 in qubole/qds-sdk-py

### LINKED ISSUE #144: Cluster actions always go through the v1.2 of the API
When creating clusters with the `v1.3` of the API we still use the `v1.2` endpoint during cluster creation. The [`_parse_create_update`](https://github.com/qubole/qds-sdk-py/blob/master/qds_sdk/cluster.py#L114) method accepts the api-version only to distinguish the request parameters. But the [`create`](https://github.com/qubole/qds-sdk-py/blob/master/qds_sdk/cluster.py#L411), [`update`](https://github.com/qubole/qds-sdk-py/blob/master/qds_sdk/cluster.py#L419) and [`clone`](https://github.com/qubole/qds-sdk-py/blob/master/qds_sdk/cluster.py#L428) methods still use the default configuration of `Qubole.agent(..)` which is set to `v1.2` unless all the actions in the session are [`configure(..)`](https://github.com/qubole/qds-sdk-py/blob/master/qds_sdk/qubole.py#L33)`ed with the v1.3 of the API. 

In my case I want to perform some of the unsupported actions with `v1.3` of the cluster API but fall back to `v1.2` for commands APIs. One way to do this is to explicitly switch the version in the `create(..)/update(..)/clone(..)` calls which thereby alter the version for that API call alone and leave the `cached_agent` as is for the v1.2 APIs.

vogxn:
When creating clusters with the v1.3 endpoint of the API we still make the request using the v1.2 endpoint during cluster creation. The `_parse_create_update` method accepts the
api-version only to distinguish the request parameters and build the request. But the create, update
and clone methods still use the default configuration of `Qubole.agent(..)` which
is set to v1.2 unless all the actions in the session are `configure(..)`ed with
the v1.3 of the API.

If you want to perform some of the unsupported actions with v1.3 of the
cluster API but fall back to v1.2 for commands APIs we explicitly switch the
version in the `create(..)/update(..)/clone(..)` calls which thereby alter the
version for that API call alone and leave the `cached_agent` as is for the v1.2
APIs.

### DIFF
--- a/qds_sdk/cluster.py
+++ b/qds_sdk/cluster.py
@@ -1165,6 +1165,7 @@ class ClusterInfoV13():
         creating or updating a cluster.
         """
         payload_dict = self.__dict__
+        payload_dict.pop("api_version", None)
         return _make_minimal(payload_dict)
 
 
--- a/qds_sdk/cluster.py
+++ b/qds_sdk/cluster.py
@@ -24,6 +24,7 @@ class Cluster(Resource):
     """
 
     rest_entity_path = "clusters"
+    api_version = "v1.2"
 
     @classmethod
     def _parse_list(cls, args):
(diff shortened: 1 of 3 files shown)


vogxn:
@prakharjain09 @vrajat

vrajat:
I understand what you are doing. Seems like a hack and we need a better long term plan. However, this diff doesnt break backward compatibility. Fine to live with the hack but it will be good to see a proper design doc to support different API versions correctly.

harshshah87:
The changes looks good.

harshshah87:
I will merge the change to unreleased today.

harshshah87:
Merged,

vogxn:
@harshshah87 I am not seeing my change in `unreleased` / `master`. Can you please check?

harshshah87:
I did in unreleased. https://github.com/qubole/qds-sdk-py/commit/b1b7e63933a87816c8a5af027d7b566c9a406aac

vogxn:
@harshshah87 that one is for the PR raised in #146 not #147

harshshah87:
Oh, yes. Let me check what happened.

harshshah87:
The change was merged to a different forked branch. Merging it again right now.

harshshah87:
Closing the pull request since its merged into unreleased and also released as part of version 1.9.1

PR #153, opened 2016-06-04
Sdk 5

harshshah87:
Support for listing the commands which are in waiting state.

### DIFF
--- a/qds_sdk/commands.py
+++ b/qds_sdk/commands.py
@@ -208,6 +208,19 @@ class Command(Resource):
             else:
                 fp.write(",".join(r['result_location']))
 
+    @classmethod
+    def get_commands_waiting(cls):
+        """
+
+        For the given API token account, return the waiting commands.
+
+        Returns:
+            Json Unicode with command information.
+
+        """
+        conn = Qubole.agent()
+        r = conn.get(cls.rest_entity_path, {"status": "waiting"})
+        return r
 
 
 class HiveCommand(Command):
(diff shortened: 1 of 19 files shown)


qds_sdk/commands.py
```diff
@@ -1155,6 +1154,60 @@ def parse(cls, args):
...
+        Returns:
+            Json Unicode with command information.
+
+        """
+        conn = Qubole.agent()
+        params = {}
+        if args["status"] is not None:
+            params["status"] = args["status"]
+        else:
+            params["status"] = "done"
```
msumit: if no status is given, then it should look for all cmd statuses?

qds_sdk/commands.py
```diff
@@ -1155,6 +1154,60 @@ def parse(cls, args):
...
+        if args["page"] is not None:
+            params["page"] = args["page"]
+        if args["per_page"] is not None:
+            params["per_page"] = args["per_page"]
```
msumit: if no `per_page` is given then how many cmds backend will send?

msumit:
Can we also use the `ListCommand` class and `list_action` method to be available with all commands types as well? e.g. `qds.py hivecmd list --status running` will show all running hive commands only.

qds_sdk/commands.py
harshshah87: per_page is 10. and page is first page by default.

qds_sdk/commands.py
harshshah87: Currently, I think the API is not working as expected and asked people to look into it. But for status waiting, it works fine. So I will hold on to the PR until it gets resolved. 
But, by default, I will do an all if it is possible.

harshshah87:
@msumit Good point. Let me check if API can work that way. If yes, I will do it.

PR #156, opened 2016-06-09
Fix scheduled command instance sdk functions.

mprice64:
This fixes a bug where when specifying page arguments to list_instances, "NameError: global name 'args' is not defined", occurs.

Also fixes a "can't combine int and str" bug with rerun


msumit:
lgtm

sourabh912:
Changes merged to unreleased branch.

PR #157, opened 2016-06-17
SDK-117: Adding live log support in command run method

msumit:
Opening up a fresh PR as prev PR (https://github.com/qubole/qds-sdk-py/pull/134/files) got all messed up.

### DIFF
--- a/qds_sdk/commands.py
+++ b/qds_sdk/commands.py
@@ -721,6 +721,7 @@ class HadoopCommand(Command):
         parsed['tags'] = options.tags
         parsed["command_type"] = "HadoopCommand"
         parsed['print_logs'] = options.print_logs
+        parsed['print_logs_live'] = options.print_logs_live
 
         if len(args) < 2:
             raise ParseError("Need at least two arguments", cls.usage)
--- a/qds_sdk/commands.py
+++ b/qds_sdk/commands.py
@@ -87,10 +87,24 @@ class Command(Resource):
         Returns:
             Command object
         """
+        err_pointer, tmp_pointer, new_bytes = 0, 0, 0
         cmd = cls.create(**kwargs)
         while not Command.is_done(cmd.status):
             time.sleep(Qubole.poll_interval)
             cmd = cls.find(cmd.id)
+            if kwargs.get('print_logs_live', False):
+                log, err_length, tmp_length = cmd.get_log_partial(err_pointer, tmp_pointer)
+
+                if err_length != "0":
+                    err_pointer += int(err_length)
+                    new_bytes = int(err_length) + int(tmp_length) - tmp_pointer
+                    tmp_pointer = int(tmp_length)
+                else:
+                    tmp_pointer += int(tmp_length)
+                    new_bytes = int(tmp_length)
+
+                if len(log) > 0 and new_bytes > 0:
+                    print >>sys.stderr, log[-new_bytes:]
 
         return cmd
 
@@ -136,6 +150,19 @@ class Command(Resource):
         r = conn.get_raw(log_path)
         return r.text
 
+    def get_log_partial(self, err_pointer=0, tmp_pointer=0):
+        """
+        Fetches log for the command represented by this object
+
+        Returns:
+            The log as a string
+        """
+        log_path = self.meta_data['logs_resource']
+        conn = Qubole.agent()
+        r = conn.get_raw(log_path, params={'err_file_processed':err_pointer, 'tmp_file_processed':tmp_pointer})
+        if 'err_length' in r.headers.keys() and 'tmp_length' in r.headers.keys():
+            return [r.text, r.headers['err_length'], r.headers['tmp_length']]
+        return [r.text, 0, 0]
 
     @classmethod
     def get_jobs_id(cls, id):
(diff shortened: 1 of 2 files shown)


msumit:
cc - @harshshah87 @adeshr @sourabh912 @yogesh2021

qds_sdk/commands.py
```diff
@@ -136,6 +150,19 @@ def get_log(self):
         r = conn.get_raw(log_path)
         return r.text
 
+    def get_log_partial(self, err_pointer=0, tmp_pointer=0):
+        """
```
sourabh912: I think instead of defining another fn i.e `get_log_partial `, can we use fn `get_log` itself ? 
Maybe modify it to something like:
 `def get_log(self, err_pointer=0, tmp_pointer=0)`
Instead of returning r.text, we now would have to return a list [] and handle it in all places where we are reading the return value from this fn.

qds_sdk/commands.py
```diff
@@ -87,10 +87,24 @@ def run(cls, **kwargs):
         Returns:
             Command object
         """
+        err_pointer, tmp_pointer, new_bytes = 0, 0, 0
         cmd = cls.create(**kwargs)
         while not Command.is_done(cmd.status):
             time.sleep(Qubole.poll_interval)
             cmd = cls.find(cmd.id)
+            if kwargs.get('print_logs_live', False):
+                log, err_length, tmp_length = cmd.get_log_partial(err_pointer, tmp_pointer)
```
sourabh912: @msumit : Can you briefly describe your logic of updating the tmp and err pointers ? 
I do not have much idea about the how we fetch the logs from tmp and error files. 
It would help me in understanding your change better.

qds_sdk/commands.py
```diff
@@ -87,10 +87,24 @@ def run(cls, **kwargs):
...
+                if err_length != "0":
+                    err_pointer += int(err_length)
+                    new_bytes = int(err_length) + int(tmp_length) - tmp_pointer
+                    tmp_pointer = int(tmp_length)
+                else:
+                    tmp_pointer += int(tmp_length)
+                    new_bytes = int(tmp_length)
+
+                if len(log) > 0 and new_bytes > 0:
+                    print >>sys.stderr, log[-new_bytes:]
```
sourabh912: Should we write the error logs to a file (default will be `sys.stderr`)? This way we can provide flexibility to the user to save logs in any file and not just `sys.stderr`.  

Please check out the comment 1 on [this](http://stackoverflow.com/questions/5574702/how-to-print-to-stderr-in-python) stackoverflow link

As a side note, `print >> sys.stderr` is not supported in Python 3.

qds_sdk/commands.py
msumit: dont want to mess around with get_log method, as it might be used by others who build their things on top of sdk

qds_sdk/commands.py
msumit: plz check old PR, there is a long discussion on that.

qds_sdk/commands.py
msumit: This is useful only when invoked via cli, so if someone wants to store logs in a file, they can call `get_log` method and redirect it to a file.

msumit: should I changed print to `sys.stderr.write(log[-new_bytes:])`?

qds_sdk/commands.py
harshshah87: Can you add this comment in the code. and edit if required. 

"""
log: Text string returned by the API.
err_length: The length of err file text (bytes).
tmp_length: The length of tmp file text which is present in the "log" var.
err_pointer: Stores the err text we have processed/recv until now and it gets passed to API in next file to indicate where to start from to get the new logs.
tmp_pointer: Same as err_pointer except it indicates the bytes of tmp file processed.

new_bytes: This is the amount of new text we need to display at every iteration.
This will be addition of the tmp_length and err_length we have recv except the tmp_pointer.
If err length is non zero, then tmp_pointer needs to be reset to the current temp_length as the err_length var will contain 
"""

qds_sdk/commands.py
harshshah87: from __future__ import print_function
 print('spam', file=sys.stderr)

This was the suggestion.

qds_sdk/commands.py
sourabh912: @msumit: `get_log` is an internal function. We should be good as long as people are only calling this function. They won't have to change signature of this fn because the extra arguments (`err_pointer `, `tmp_pointer `) that we want to add, will have default values.

msumit: nop, when someone imports sdk, they access this method directly. we can add extra args without breaking them, but response is changing as well here, which might break existing tools.

qds_sdk/commands.py
msumit: added some comments

qds_sdk/commands.py
msumit: added python3 compatible print

qds_sdk/commands.py
```diff
@@ -87,10 +88,28 @@ def run(cls, **kwargs):
         Returns:
             Command object
         """
+
+        # vars to keep track of actual logs bytes (err, tmp) and new bytes seen in each iteration
+        err_pointer, tmp_pointer, new_bytes = 0, 0, 0
         cmd = cls.create(**kwargs)
```
harshshah87: One last thing from my side.

Can you pop "print-live-logs" from the kwargs here. We are not doing it during runaction and sending this argument to the API. To be consistent in the print-logs popping which happened in runaction, you can use positional argument to be passed here or simply pop before create like we do it in submitaction(). based on what approach you feel is the best, keep the '''print_logs = args.pop("print_logs")''' line from runaction together here. I.e like submitaction, before issuing a create, pop all arguments not required and later use them. 

I think while testing it, API must have ignored the extra argument. 

There are two review comments which looks active (including this one) and once we have closure on these, I can merge the changes over the weekend.

msumit: done

harshshah87:
Merged the changes:
Commit id: 83ec6b30862843acb38d966687111a89d15414ab

PR #159, opened 2016-06-24
Sdk for branding and adding link to sub accounts

tanishgupta1:
This sdk will enable logo feature and links for other customers within user account.

### DIFF
--- a/qds_sdk/account.py
+++ b/qds_sdk/account.py
@@ -76,7 +76,6 @@ class AccountCmdLine:
     @staticmethod
     def branding(args):
         v= {}
-        print(args)
         v['account_id'] = args.account_id
         if args.logo_uri is not None:
             v['logo'] = {'logo_uri' : args.logo_uri }
--- a/qds_sdk/account.py
+++ b/qds_sdk/account.py
@@ -43,6 +43,15 @@ class AccountCmdLine:
             choices=["true", "false"], default="false",
             help="Use previous account plan, default: false")
         create.set_defaults(func=AccountCmdLine.create)
+
+        #branding
+        branding = subparsers.add_parser(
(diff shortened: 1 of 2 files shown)


qds_sdk/account.py
```diff
@@ -64,6 +73,25 @@ def create(args):
         result = Account.create(**v)
         return result
 
+    @staticmethod
+    def branding(args):
+        v= {}
+        print(args)
```
harshshah87: Remove the debugging statement.

qds_sdk/account.py
```diff
@@ -43,6 +43,15 @@ def parsers():
             choices=["true", "false"], default="false",
             help="Use previous account plan, default: false")
         create.set_defaults(func=AccountCmdLine.create)
+
+        #branding
+        branding = subparsers.add_parser(
```
harshshah87: Confirmation from API perspective.
1. This API overwrites the logo--uri in case there already exist one for a given account? 
2. All the three links/uri are optional. Making a blank call with just account-id will have no effect. 
3. What is the label of documentation URL?

tanishgupta1: 1. yes in api it is already checking if uri exist or not for a given account.
2.making a blank call with just account id will return status "Enter either logo or link parameters" from api.
3. label of documentation url is name of the link which user want to give.Should I change it to name of link-url ?

harshshah87: 3. I checked the controller, since even this one uses link-url, we are fine. 

Will merge the changes today.

qds_sdk/account.py
tanishgupta1: done

harshshah87:
Merged. Closing the PR.

PR #163, opened 2016-06-29
SDK-124: Fixing KeyError when run is called by code and not via cli

msumit:


harshshah87:
Merging the changes and releasing the new one.

harshshah87:
Merging the changes.

PR #167, opened 2016-08-02
Object acl changes

harshshah87:
Need an API support to set/get policy for resource level.

### DIFF
--- a/bin/qds.py
+++ b/bin/qds.py
@@ -66,7 +66,7 @@ usage_str = (
     "    update_node: update a node on a existing cluster\n"
     "    get_snapshot_schedule: get details of scheduled snapshots on a hbase cluster\n"
     "    update_snapshot_schedule: update scheduled snapshots on a hbase cluster\n"
-    "    object_policy: View/Update the access conditions for a cluster.\n"
+    "    object_policy: View/Update the access policy of a cluster.\n"
     "\nDbTap subcommand:\n"
     "  dbtap --help\n"
     "\nReport subcommand:\n"
--- a/bin/qds.py
+++ b/bin/qds.py
@@ -16,7 +16,8 @@ from qds_sdk.app import AppCmdLine
 from qds_sdk.nezha import NezhaCmdLine
 from qds_sdk.user import UserCmdLine
 from qds_sdk.template import TemplateCmdLine
-from qds_sdk.object_policy import ObjectPolicy
+from qds_sdk.object_policy import ObjectPolicyCmdLine
+from qds_sdk.cluster import ClusterObjectPolicyCmdLine
 
 import os
 import sys
@@ -373,8 +374,8 @@ def cluster_status_action(clusterclass, args):
     return 0
 
 def cluster_object_policy_action(clusterclass, args):
-    arguments = ObjectPolicy._parse_object_policy(args, "object_policy", clusterclass.__name__)
-    result = ObjectPolicy.process_object_policy(arguments)
+    arguments = ClusterObjectPolicyCmdLine._parse_object_policy(args, "object_policy", clusterclass.__name__)
+    result = ClusterObjectPolicyCmdLine.process_object_policy(arguments)
     print (json.dumps(result, indent=4))
 
 def cluster_reassign_label_action(clusterclass, args):
--- a/bin/qds.py
+++ b/bin/qds.py
@@ -65,6 +65,8 @@ usage_str = (
     "    update_node: update a node on a existing cluster\n"
     "    get_snapshot_schedule: get details of scheduled snapshots on a hbase cluster\n"
     "    update_snapshot_schedule: update scheduled snapshots on a hbase cluster\n"
+    "    get_access_control: View the access conditions for a cluster.\n"
+    "    set_access_control: Update the access conditions for a cluster.\n"
     "\nDbTap subcommand:\n"
     "  dbtap --help\n"
     "\nReport subcommand:\n"
@@ -370,6 +372,15 @@ def cluster_status_action(clusterclass, args):
     print(json.dumps(result, indent=4))
     return 0
 
+def cluster_get_access_control_action(clusterclass, args):
+    arguments = clusterclass._parse_access_control(args,"get_access_policy")
+    result = clusterclass.get_access_control(arguments.cluster_id)
+    print (json.dumps(result, indent=4))
+
+def cluster_set_access_control_action(clusterclass, args):
+    arguments = clusterclass._parse_access_control(args, "set_access_policy")
+    result = clusterclass.set_access_control(arguments.cluster_id, arguments.policy)
+    print (json.dumps(result, indent=4))
 
 def cluster_reassign_label_action(clusterclass, args):
     arguments = clusterclass._parse_reassign_label(args)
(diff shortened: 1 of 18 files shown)


bin/qds.py
```diff
@@ -65,6 +66,7 @@
     "    update_node: update a node on a existing cluster\n"
     "    get_snapshot_schedule: get details of scheduled snapshots on a hbase cluster\n"
     "    update_snapshot_schedule: update scheduled snapshots on a hbase cluster\n"
+    "    object_policy: View/Update the access conditions for a cluster.\n"
```
msumit: `conditions for` or `policies of`?

msumit:
Can you paste examples of using OACLs via cli and through a python script?

msumit:
You also have to add specs for `object_policy` model and its interaction via `clusters`..

harshshah87:
Command line usage:


# Update a policy:
qds.py --url="https://api.qubole.net/api" --token="$API_TOKEN" cluster object_policy update --cluster-id 201 --policy "[{\"access\": \"allow\", \"action\": [\"all\"], \"condition\": {\"qbol_users\": [55]} }]"
{
    "message": "Policy saved successfully.",
    "id": 2,
    "success": true
}

#get a policy
qds.py --url="https://api.qubole.net/api" --token=“$AUTH_TOKEN” cluster object_policy get --cluster-id 201
{
    "policy": [
        {
            "access": "allow",
            "action": [
                "all"
            ],
            "condition": {
                "qbol_users": [
                    55
                ]
            }
        }
    ],
    "id": 1
}



Sample python script to update and get a policy. Output will be similar to the one above.
========================================================================
from qds_sdk.object_policy import ObjectPolicy
from qds_sdk.cluster import Cluster
from qds_sdk.qubole import Qubole
import json

token = “YOUR_TOKEN”
Qubole.configure(api_token=token, version="latest", api_url="https://api.qubole.net/api")
policy = """  [{\"access\": \"allow\", \"action\": [\"all\"], \"condition\": {\"qbol_users\": [55]} }]  """
print ObjectPolicy.update_object_policy(201, Cluster.__name__, policy)
print ObjectPolicy.get_object_policy(201, Cluster.__name__)
========================================================================

qds_sdk/object_policy.py
```diff
@@ -0,0 +1,69 @@
...
+        update_parser.add_argument("--policy", required = True,
+                                    help="Policy Statement example '[{\"access\":\"deny\", \"action\":"
+                                         " [\"all\"], \"condition\": {\"qbol_users\": [user_id,], \"qbol_groups\": [group_ids,]} }]")
+        arguments = argsParser.parse_args(args)
+        return arguments
+
+    @classmethod
+    def process_object_policy(cls, arguments):
+        if arguments.cmd == "get":
+            return cls.get_object_policy( arguments.id, arguments.source_type)
```
msumit: lets switch the agrs position, `source_type` should come before then `source_id`

qds_sdk/object_policy.py
```diff
@@ -0,0 +1,69 @@
...
+
+    @classmethod
+    def _parse_object_policy(cls, args, action, object):
+        """
+        Args:
+            `args`: sequence of arguments
+            `action` : "get_object_policy" or "update_object_policy"
+
+        Returns:
+            Object that contain cluster parameters
```
msumit: does it return `cluster` parameters? or policy information?

msumit:
@harshshah87 can u paste the output of `help` text when user interacts on `cli`? I mean this `qds.py  --token="$API_TOKEN" cluster object_policy --help`

harshshah87:
qds.py --url="https://larva.qubole.net/api" --token="$token" cluster object_policy
usage: Cluster object_policy [-h] {get,update} ...

positional arguments:
  {get,update}
optional arguments:
  -h, --help    show this help message and exit

qds_sdk/object_policy.py
harshshah87: Should be a dictionary which has parsed command line argument.

msumit:
lgtm.. but we need some more eyes.. @vrajat @itsvikramagr

tests/test_cluster.py
```diff
@@ -2285,5 +2285,22 @@ def test_update(self):
         with self.assertRaises(SystemExit):
             qds.main()
 
+class TestClusterACL(QdsCliTestCase):
+    def test_get_object_policy(self):
+        sys.argv = ['qds.py', 'cluster', 'object_policy', 'get', '--cluster-id', '1001']
+        print_command()
+        Connection._api_call = Mock(return_value={})
+        qds.main()
+        Connection._api_call.assert_called_with('GET', 'object_policy/get_object_policy', params ={'source_id':'1001',
```
vrajat: I dont this will work. `get_object_policy` is not required in the URL. Have you tested it for real ? If it does then the API design is wrong. It should just be `GET object_policy`

tests/test_cluster.py
```diff
@@ -2285,5 +2285,22 @@ def test_update(self):
         with self.assertRaises(SystemExit):
             qds.main()
 
+class TestClusterACL(QdsCliTestCase):
+    def test_get_object_policy(self):
+        sys.argv = ['qds.py', 'cluster', 'object_policy', 'get', '--cluster-id', '1001']
```
vrajat: why is `cluster` required in `--cluster-id` ? You've already specified `cluster` why do it twice ? Also I dont like the way the command line is setup. Maybe the recommendation is completely broken across the SDK. However, the preferred pattern is 

    qds.py object action [args]


With that pattern, it should be

    qds.py cluster get_object_policy --id

vrajat:
I prefer different design. I dont know enough about `ObjectPolicy`. I am assuming that different objects support policy and therefore you have tried to abstract it out. `Cluster` class is not the best designed. Check out `Scheduler` or `User` etc. They are split into `Scheduler` and `SchedulerCmdLine` or `User` and `UserCmdLine`. My point is that there is no use trying to abstract out this functionality for a class like 'Cluster`. 

Instead what I suggest is you put the code for cluster into cluster and not put it in a different class. 
Then, you create a `PolicyCmdLine` and `PolicyResource(Resource)` base classes which well-designed classes like `Scheduler` or `User` can inherit from. For e.g. their definition will look like:

    SchedulerCmdLine(PolicyCmdLine)
    ...
    ...

    Scheduler(PolicyResource)
    ...
    ...

tests/test_cluster.py
harshshah87: get_object_policy is required. and just "GET  object_policy" returns 500 error. @msumit

tests/test_cluster.py
harshshah87: I can make it --id. It makes more sense too.

tests/test_cluster.py
msumit: `get/update/delete_object_policy` methods are for UI, as UI didnt wanted to do 2 calls for a proper CRUD. We do've proper CRUD apis available for Object_policy as well. So idle calls chain would be like `GET cluster/:cid/object_policy` and then `GET object_policy/:opid`.. but as we didn't added those `get object_policy` methods to all the resources, what @harshshah87 did is fine.

harshshah87:
@vrajat I am still thinking into why do we definitely need two different classes. Is the reason that command line code should be different as they are not supposed to be used by the sdk.

If i can improvise the suggestion a little more, we want the base class Resource class to support policy by default (this is instead of keeping a middle hop like PolicyResource)?

The way I saw it is object policy is a different API exposed and we should have a different class. For command line, it parses based on the subcommand which will be treated as object.

tests/test_cluster.py
vrajat: IMO, SDK should use proper REST calls. Not UI calls. There should be a very good reason to break design patterns.

vrajat:
You should extend `Resource` class only if all objects are ready to support the policy object. If thats true, then its fine. However, I am assuming not all the objects support it. 

Re: different object. I am not sure of the API design. Should it have been nested into each object or should it be first level with source id/source_type ? I wasnt part of the discussions.

However, the command line you have designed is completely different. Instead it should have been:

    qds.py policy show --type cluster --id <id>

harshshah87:
object_policy will be part of clusters, and hence it needs to be accessed using the clusters command line. So we are not going to have a separate command line for policy. I think UI also does the same thing where you go to the policy via the clusters.

So the ask here is, clusters should have the update and get API exposed and the command line for clusters should be parsing the arguments. Is this the recommendation?  If we do both things, there will be some code duplication (each object parser will need to support the command line arguments). I will need to find a way to avoid it then.

vrajat:
The way to avoid code duplication is to have PolicyResource and
PolicyCmdLine which can be inherited by others. But hard-code it for
clusters because clusters unfortunately does not follow the
Resource/CmdLine pattern.

Also you should not be using UI end points but REST end points.

PR #176, opened 2016-11-10
Doc:updating doc for macros

akhileshmaurya:


harshshah87:
We do not want to update the help file with very specific example. The command line help will be available for this: `qds.py --token=$TOKEN hivecmd run help`

PR #179, opened 2016-12-02
Integrating the new api call to get account credentials

anumahadevan:
Modified the credentials_rest_entity_path in accounts class to adhere to the new Api.(https://azureqa.qubole.net/api/v1.2/accounts/creds). Also fixed small copy paste issues in set_cloud_config function for ClusterInfoV2 class and also removed the overrided set_node_config temporarily. The set_node_config should be implemented again as part of taking in the v2 api changes.

### DIFF
--- a/qds_sdk/account.py
+++ b/qds_sdk/account.py
@@ -94,7 +94,7 @@ class AccountCmdLine:
         return result
 
 class Account(SingletonResource):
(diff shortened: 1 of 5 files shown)


sourabh912:
@anumahadevan : Looks like the tests are failing. Can you fix them ?

PR #180, opened 2016-12-02
Integrating the new account creds api for getting account information

anumahadevan:
Adding the new account credentials api - https://qa3.qubole.net/api/v1.2/accounts/creds instead of get_creds api call.


qds_sdk/commands.py
```diff
@@ -228,9 +228,9 @@ def get_results(self, fp=sys.stdout, inline=True, delim=None, fetch=True):
         else:
             if fetch:
                 storage_credentials = conn.get(Account.credentials_rest_entity_path)
-                boto_conn = boto.connect_s3(aws_access_key_id=storage_credentials['storage_access_key'],
-                                            aws_secret_access_key=storage_credentials['storage_secret_key'],
-                                            security_token = storage_credentials['session_token'])
+                boto_conn = boto.connect_s3(aws_access_key_id=storage_credentials.get('storage_config').get('access_key'),
+                                            aws_secret_access_key=storage_credentials.get('storage_config').get('secret_key'),
+                                            security_token = storage_credentials.get('storage_config').get('session_token'))
```
tanishgupta1: Is this the v2 version of api we are using for getting account credentials ?

anumahadevan: No this is the v1.2 version new account API that got added, https://qa3.qubole.net/api/v1.2/accounts/creds

tanishgupta1: Its confusing that we are using new v2 response of api but we are supporting only v1.2 as version. But since we don't support multiple version in account , this looks fine. We can support multiple version change later.

PR #182, opened 2016-12-14
Clusterinfo v2 payload changes and adding the new credentials api

anumahadevan:
Integrating the multicloud supported v2 api payload in the clusterinfov2. Also added changes for new account credentials api and instance info api.


qds_sdk/cluster.py
```diff
@@ -1168,6 +1214,217 @@ def minimal_payload(self):
...
+    def set_monitoring(self, ganglia=None, datadog_api_token=None, datadog_app_token=None):
+        self.monitoring['ganglia'] = ganglia
+        self.set_datadog_settings(datadog_api_token, datadog_app_token)
+
+    def set_datadog_settings(self, datadog_api_token=None, datadog_app_token=None):
+        self.monitoring['datadog'] = {}
+        self.monitoring['datadog']['datadog_api_token'] = datadog_api_token
+        self.monitoring['datadog']['datadog_app_token'] = datadog_app_token
+
+    def set_internal(self, spark_s3_package_name=None, zeppelin_s3_package_name=None, zeppelin_interpreter_mode=None,
```
vogxn: Suggest removing the internal section from being edited by the SDK. Ideally we leave debug hooks here that should not be exposed to typical users of the SDK. WDYT?

qds_sdk/cluster.py
```diff
@@ -1168,6 +1214,217 @@ def minimal_payload(self):
...
+        self.engine_config['spark_settings']['spark_version'] = spark_version
+        self.engine_config['spark_settings']['custom_spark_config'] = custom_spark_config
+
+    def set_airflow_settings(self, dbtap_id=None, fernet_key=None, overrides=None):
+        self.engine_config['airflow_settings'] = {}
+        self.engine_config['airflow_settings']['dbtap_id'] = dbtap_id
+        self.engine_config['airflow_settings']['fernet_key'] = fernet_key
+        self.engine_config['airflow_settings']['overrides'] = overrides
+
+    def set_monitoring(self, ganglia=None, datadog_api_token=None, datadog_app_token=None):
```
vogxn: If the ganglia option is a boolean, let's call it `enable_ganglia`. Conveys what argument should be passed correctly.

qds_sdk/cluster.py
```diff
@@ -1168,6 +1214,217 @@ def minimal_payload(self):
...
+            except Exception as e:
+                raise Exception("Invalid JSON string for custom ec2 tags: %s" % e.message)
+        self.set_data_disk(size, count, type, upscaling_config, encryption)
+        self.cluster_info['heterogeneous_config'] = heterogeneous_config
+        self.cluster_info['slave_request_type'] = slave_request_type
+        self.set_spot_settings(maximum_bid_price_percentage, timeout_for_request,
+                                        maximum_spot_instance_percentage, stable_maximum_bid_price_percentage,
+                                        stable_timeout_for_request, stable_allow_fallback)
+
+    def set_spot_settings(self, maximum_bid_price_percentage=None, timeout_for_request=None,
```
vogxn: If I call `set_spot_settings(..)` followed by `set_cluster_information(..)` without the spot settings specified. We'll end up overwriting, correct? Is that desirable? Similar problems will happen when you make a member method of this class implicitly from within another member method that takes a super set of arguments.

qds_sdk/cluster.py
```diff
@@ -1168,6 +1214,217 @@ def minimal_payload(self):
...
+    def set_stable_spot_bid_settings(self, stable_maximum_bid_price_percentage=None,
+                          stable_timeout_for_request=None, stable_allow_fallback=None):
+        self.cluster_info['spot_settings']['stable_spot_bid_settings'] = {}
+        self.cluster_info['spot_settings']['stable_spot_bid_settings']['stable_maximum_bid_price_percentage'] = stable_maximum_bid_price_percentage
+        self.cluster_info['spot_settings']['stable_spot_bid_settings']['stable_timeout_for_request'] = stable_timeout_for_request
+        self.cluster_info['spot_settings']['stable_spot_bid_settings']['stable_allow_fallback'] = stable_allow_fallback
+
+    def set_data_disk(self, size=None, count=None, type=None, upscaling_config=None, encryption=None):
+        self.cluster_info['datadisk'] = {}
+        self.cluster_info['datadisk']['size'] = size
```
vogxn: Should `size` and `count` be integers?

qds_sdk/cluster.py
```diff
@@ -1168,6 +1214,217 @@ def minimal_payload(self):
...
+        self.cluster_info['datadisk']['count'] = count
+        self.cluster_info['datadisk']['type'] = type
```
vogxn: Don't use `type` as a field name. That shadows the built-in Python `type` and can undesirable effects.

qds_sdk/cluster.py
```diff
@@ -1168,6 +1214,217 @@ def minimal_payload(self):
...
+        self.cluster_info['spot_settings']['spot_instance_settings']['maximum_spot_instance_percentage'] = maximum_spot_instance_percentage
+
+    def set_stable_spot_bid_settings(self, stable_maximum_bid_price_percentage=None,
+                          stable_timeout_for_request=None, stable_allow_fallback=None):
+        self.cluster_info['spot_settings']['stable_spot_bid_settings'] = {}
+        self.cluster_info['spot_settings']['stable_spot_bid_settings']['stable_maximum_bid_price_percentage'] = stable_maximum_bid_price_percentage
+        self.cluster_info['spot_settings']['stable_spot_bid_settings']['stable_timeout_for_request'] = stable_timeout_for_request
+        self.cluster_info['spot_settings']['stable_spot_bid_settings']['stable_allow_fallback'] = stable_allow_fallback
+
+    def set_data_disk(self, size=None, count=None, type=None, upscaling_config=None, encryption=None):
```
vogxn: Encryption is a boolean right? Argument name should be changed

qds_sdk/cluster.py
```diff
@@ -24,7 +24,7 @@ class Cluster(Resource):
     """
 
     rest_entity_path = "clusters"
-    api_version = "v1.2"
+    api_version = "v1.3"
```
tanishgupta1: Is this change necessary ?
If the default was v1.2 and we changed it to v1.3 then the apis which are written by customers might fail.

vogxn: I am pretty sure it is impossible to use the cluster API as it is exposed from the SDK today. There are way too many options to setup. This change overall affects only the quality team within Qubole. :)

qds_sdk/cluster.py
anumahadevan: Yes thanks, will fix that

qds_sdk/cluster.py
tanishgupta1: Just for knowledge. So this change will go to unreleased branch and later on this will be merge to master. So for unreleased branch it will affect quality team only but once it is merge to master it will be used by everyone ?

vogxn: @anumahadevan this will cause the builds of our CI to also use the unreleased branch. And since the tests currently run from master branch this could potentially cause problems? Since the test factory work is still unmerged.

sourabh912:
@anumahadevan : There are no unit tests for v2 apis. Would be great if you can add them.

qds_sdk/cluster.py
sourabh912: If we are changing the default version to 1.3, then we would need to make changes at other places as well for example in `qds.py`, `qubole.py`. Also we need to update comments at various places in the code.

qds_sdk/cluster.py
```diff
@@ -24,7 +24,7 @@ class Cluster(Resource):
     """
 
     rest_entity_path = "clusters"
-    api_version = "v1.2"
```
sourabh912: This api_version is useless I guess. 

We pass the appropriate version from `qds.py`.

qds_sdk/cluster.py
```diff
@@ -66,9 +66,10 @@ def list(cls, state=None):
         Returns:
             List of clusters satisfying the given criteria
         """
-        conn = Qubole.agent()
+        conn = Qubole.agent(version=Cluster.api_version)
```
sourabh912: If we change the default version to v1.3 at all places, then we would not need this change. 
Also I think we pass the version info to various fns of class `Cluster` via qds.py. For example check `update()`, `clone()`

qds_sdk/cluster.py
sourabh912: @anumahadevan : I would suggest to not change the default api version in this PR unless there is a strong reason to do so. 
We can do the version change in a separate PR.

qds_sdk/account.py
```diff
@@ -94,7 +94,7 @@ def branding(args):
         return result
 
 class Account(SingletonResource):
-    credentials_rest_entity_path = "accounts/get_creds"
+    credentials_rest_entity_path = "accounts/creds"
```
sourabh912: What is the motivation for this change ? 
It this change is irrelevant to this PR, then lets take it up in a separate PR.

qds_sdk/cluster.py
```diff
@@ -383,27 +310,276 @@ def _parse_create_update(cls, args, action, api_version):
...
+                                                    default=None,
+                                                    help="Fallback to on-demand nodes if spot nodes" +
+                                                         " could not be obtained. Valid only if slave_request_type is spot", )
+            fallback_to_ondemand_group.add_argument("--no-fallback-to-ondemand",
+                                                    dest="fallback_to_ondemand",
+                                                    action="store_false",
+                                                    default=None,
+                                                    help="Dont Fallback to on-demand nodes if spot nodes" +
+                                                         " could not be obtained. Valid only if slave_request_type is spot", )
+            if api_version == 1.3:
```
sourabh912: Why are we checking the api version here since we are already inside `api_version == 1.3` (Line: 328)

qds_sdk/cluster.py
```diff
@@ -1169,6 +1350,324 @@ def minimal_payload(self):
...
+                raise Exception("Invalid JSON string for custom ec2 tags: %s" % e.message)
+
+        self.cluster_info['heterogeneous_config'] = heterogeneous_config
+        self.cluster_info['slave_request_type'] = slave_request_type
+
+
+    def set_spot_instance_settings(self, maximum_bid_price_percentage=100, timeout_for_request=10,
+                          maximum_spot_instance_percentage=50):
+        self.cluster_info['spot_settings']['spot_instance_settings'] = {}
+        self.cluster_info['spot_settings']['spot_instance_settings']['maximum_bid_price_percentage'] = maximum_bid_price_percentage
```
sourabh912: nit: Please split these statements in two lines.

qds_sdk/account.py
anumahadevan: This is relevant to the multicloud initiative. The get_creds api is replaced by creds api as a part of multicloud efforts.

qds_sdk/cluster.py
anumahadevan: We need these statements only for 1.3 version. We are inside the if statement which checks api_version >= 1.3 which includes v2.0 also.

qds_sdk/cluster.py
anumahadevan: Ok sure.

qds_sdk/commands.py
```diff
@@ -228,9 +228,9 @@ def get_results(self, fp=sys.stdout, inline=True, delim=None, fetch=True):
         else:
             if fetch:
                 storage_credentials = conn.get(Account.credentials_rest_entity_path)
-                boto_conn = boto.connect_s3(aws_access_key_id=storage_credentials['storage_access_key'],
-                                            aws_secret_access_key=storage_credentials['storage_secret_key'],
-                                            security_token = storage_credentials['session_token'])
+                boto_conn = boto.connect_s3(aws_access_key_id=storage_credentials.get('storage_config').get('access_key'),
```
abhisheksr8: @anumahadevan this wont work with v1.2 and v1.3, we should make it backward compatible

qds_sdk/account.py
abhisheksr8: @anumahadevan this fails if user tries v1.2, also tried this with v2 and is failing at this endpoint

qds_sdk/commands.py
anumahadevan: Account apis still use only v1.2. So here we are making use of the more cloud agnostic API (/creds) instead of get_creds API. The api change will be pushed into production with r40. It might not work with api.qubole.com now.

sourabh912:
@anumahadevan: I have merged this PR into `multicloud` branch in Qubole's repo.

PR #183, opened 2016-12-22
Added include_headers param in get_results method

tanishgupta1:
This param will add headers on results if its true.

### DIFF
--- a/bin/qds.py
+++ b/bin/qds.py
@@ -45,8 +45,8 @@ usage_str = (
     "    run [cmd-specific-args .. ] : submit cmd & wait. print results\n"
     "    check <id> : print the cmd object for this id\n"
     "    cancel <id> : cancels the cmd with this id\n"
-    "    getresult <id> <enable_header>: id -> get the results for the cmd with this id\n"
-    "                                    enable_header -> to include headers in results(true/false)\n"
+    "    getresult <id> <include_header>: id -> get the results for the cmd with this id\n"
+    "                                     include_header -> to include headers in results(true/false)\n"
     "    getlog <id> : get the logs for the cmd with this id\n"
     "\nCluster subcommand:\n"
     "  cluster <action>\n"
@@ -117,7 +117,7 @@ def submitaction(cmdclass, args):
 def _getresult(cmdclass, cmd, args=[]):
     if Command.is_success(cmd.status):
         log.info("Fetching results for %s, Id: %s" % (cmdclass.__name__, cmd.id))
-        cmd.get_results(sys.stdout, delim='\t', arguments=args)
+        cmd.get_results(sys.stdout, delim='\t', qlog=cmd.qlog, arguments=args)
         return 0
     else:
         log.error("Cannot fetch results - command Id: %s failed with status: %s" % (cmd.id, cmd.status))
--- a/bin/qds.py
+++ b/bin/qds.py
@@ -16,6 +16,7 @@ from qds_sdk.app import AppCmdLine
 from qds_sdk.nezha import NezhaCmdLine
 from qds_sdk.user import UserCmdLine
 from qds_sdk.template import TemplateCmdLine
+from qds_sdk.sensors import *
 
 import os
 import sys
@@ -35,6 +36,11 @@ CommandClasses = {
     "dbexportcmd": DbExportCommand,
     "dbimportcmd": DbImportCommand,
     "prestocmd": PrestoCommand
+}
+
+SensorClasses = {
+    "filesensor": FileSensor,
+    "partitionsensor": PartitionSensor
 }
 
 usage_str = (
@@ -87,7 +93,9 @@ usage_str = (
     "\nNezha subcommand:\n"
     "  nezha --help\n"
     "\nUser subcommad:\n"
-    "  user --help\n")
+    "  user --help\n"
+    "\nSensor subcommand:\n"
+    " <filesensor|partitionsensor> --help\n")
 
 
 def usage(parser=None):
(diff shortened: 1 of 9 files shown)


bin/qds.py
```diff
@@ -45,7 +45,8 @@
     "    run [cmd-specific-args .. ] : submit cmd & wait. print results\n"
     "    check <id> : print the cmd object for this id\n"
     "    cancel <id> : cancels the cmd with this id\n"
-    "    getresult <id> : get the results for the cmd with this id\n"
+    "    getresult <id> <enable_header>: id -> get the results for the cmd with this id\n"
+    "                                    enable_header -> to include headers in results(true/false)\n"
     "    getlog <id> : get the logs for the cmd with this id\n"
```
msumit: `enable_headers` doesn't sound correct.. u can keep it `include_headers` only..

bin/qds.py
```diff
@@ -113,10 +114,10 @@ def submitaction(cmdclass, args):
         return 0
 
 
-def _getresult(cmdclass, cmd):
+def _getresult(cmdclass, cmd, args=[]):
```
msumit: use `kwargs` ie a hash instead of an array..

bin/qds.py
```diff
@@ -156,9 +157,12 @@ def cancelaction(cmdclass, args):
 
 
 def getresultaction(cmdclass, args):
-    checkargs_id(args)
+    if len(args) > 2:
+        sys.stderr.write("expecting not more than 2 arguments\n")
+        usage()
+
```
msumit: no need of this check, once u make args a hash

qds_sdk/commands.py
```diff
@@ -212,7 +212,11 @@ def get_results(self, fp=sys.stdout, inline=True, delim=None, fetch=True):
 
         conn = Qubole.agent()
 
-        r = conn.get(result_path, {'inline': inline})
+        enable_header = "false"
+        if len(arguments) == 1:
+            enable_header = arguments.pop(0)
```
msumit: no a good way to accept parameters.. make args a hash..

tests/test_command.py
```diff
@@ -5,7 +5,7 @@
     import unittest
 else:
     import unittest2 as unittest
-from mock import Mock
+from mock import *
```
msumit: why this change needed?

tanishgupta1: call function was not declared which is present in mock , so included * to solve all dependencies. Same thing is done on test_actions.py to solve this.

bin/qds.py
tanishgupta1: Looked into it.We can't pass hash here as its a single command line argument although it can be converted to hash internally but then we have create new function by converting single line arguments to hash . We generally use hash where we we use optparser or argumentparser and in return we get hash which is not the case here.

bin/qds.py
tanishgupta1: same as above.

qds_sdk/commands.py
tanishgupta1: same as above.

msumit:
lgtm

qds_sdk/cluster.py
```diff
@@ -737,6 +744,7 @@ def set_ec2_settings(self,
         self.ec2_settings['vpc_id'] = vpc_id
         self.ec2_settings['subnet_id'] = subnet_id
         self.ec2_settings['role_instance_profile'] = role_instance_profile
+        self.ec2_settings['master_elastic_ip'] = master_elastic_ip
```
harshshah87: NIT: Can you add the setting description to the above Kwargs text.

qds_sdk/commands.py
```diff
@@ -212,7 +212,11 @@ def get_results(self, fp=sys.stdout, inline=True, delim=None, fetch=True):
 
         conn = Qubole.agent()
 
-        r = conn.get(result_path, {'inline': inline})
+        include_header = "false"
+        if len(arguments) == 1:
+            include_header = arguments.pop(0)
```
harshshah87: Can we have a validation here= > include_header in ["true","false"]

bin/qds.py
```diff
@@ -113,10 +114,10 @@ def submitaction(cmdclass, args):
         return 0
 
 
-def _getresult(cmdclass, cmd):
+def _getresult(cmdclass, cmd, args=[]):
     if Command.is_success(cmd.status):
         log.info("Fetching results for %s, Id: %s" % (cmdclass.__name__, cmd.id))
-        cmd.get_results(sys.stdout, delim='\t')
+        cmd.get_results(sys.stdout, delim='\t', qlog=cmd.qlog, arguments=args)
```
tanishgupta1: cmd is the object which is returned when we run the command. So qlog is one of the attribute of that object same like cmd.status in above condition.

qds_sdk/cluster.py
tanishgupta1: I think this is not my change. Have to merge latest unreleased branch to this.

qds_sdk/commands.py
```diff
@@ -1065,6 +1076,8 @@ class DbImportCommand(Command):
                          help="Can be 1 for Hive export or 2 for HDFS/S3 export")
     optparser.add_option("--hive_table", dest="hive_table",
                          help="Mode 1: Name of the Hive Table from which data will be exported")
+    optparser.add_option("--hive_serde", dest="hive_serde",
```
tanishgupta1: same above this is not my change. Its due to merge issue.

qds_sdk/commands.py
tanishgupta1: done

PR #187, opened 2017-01-30
SDK-178: Adding File & Partition Sensors

msumit:
Note: Sensors api will be in prod with rb41.

### DIFF
--- a/qds_sdk/sensors.py
+++ b/qds_sdk/sensors.py
@@ -14,6 +14,33 @@ import json
 log = logging.getLogger("qds_sensors")
 
 
+class SensorCmdLine:
+
+    @staticmethod
+    def check(cls, args):
+        """
+        Method to call Sensor.check after parsing args from cmdline
+        :param cls: sensor class
+        :param args: inline arguments
+        :return: True or False
+        """
+        parser = SensorCmdLine.parsers(cls)
+        parsed = parser.parse_args(args)
+        return cls.check(json.loads(parsed.data))
+
+    @staticmethod
+    def parsers(cls):
+        argparser = ArgumentParser(prog=cls.usage, description=cls.description)
+        subparsers = argparser.add_subparsers()
+
+        #Check
+        check = subparsers.add_parser("check", help="Check a Sensor")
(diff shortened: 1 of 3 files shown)


yogesh2021:
A qubole operator change will also follow up with this ?

msumit:
@yogesh2021 yes there is PR opened in Airflow repo (bitbucket) as well..

qds_sdk/sensors.py
```diff
@@ -0,0 +1,71 @@
...
+        Method to call check after parsing args from cli
+        :param args: inline arguments
+        :return: True or False
+        """
+        parser = cls.parsers()
+        parsed = parser.parse_args(args)
+        return cls.check(json.loads(parsed.data))
+
+    @classmethod
+    def parsers(cls):
```
harshshah87: The function parsers and check_cli(we can rename it to check) should be in new SensorCmdLine class. The dependency on usage and description from the derived classes can be removed.
usage = "qds.py %s check -d 'json string'" % cls.__name__.lower()
description (do we really need one) can be set in similar manner.

msumit: Any specific reason to move these methods into new `SensorCmdLine` class? I followed the design of `Command` class here.. Also I think `cls` variable won't be available in parent class, as its not inside a class method?

msumit: @harshshah87 moved `check` and `parser` to `SesnorCmdLine` class..

qds_sdk/sensors.py
```diff
@@ -0,0 +1,75 @@
...
+import logging
+import json
+
+log = logging.getLogger("qds_sensors")
+
+
+class SensorCmdLine:
+
+    @staticmethod
+    def check(cls, args):
```
harshshah87: NIT: Need to rename the variable from cls to sensor_class. The \@staticmethod along with cls as variable name is confusing.

msumit:
@harshshah87 done.. can u plz squash and merge?

harshshah87:
Merged to unreleased.

PR #189, opened 2017-04-12
MW-172-1: Cloud Agnostic changes for v2 api cluster requests

tanishgupta1:
* Added cloud as extra variable for qubole configuration setting.

* Created aws , oracle, azure classes to manage cloud config setting, argparser of cluster or other cloud specific functions.

* Added engine class to manage engine specific functions.

* Manage command line arguments accordingly w.r.t cloud. So for aws cloud only aws specific command line arguments will be shown. Similarly for azure and oracle.


tanishgupta1:
* Added cloud as extra variable for qubole configuration setting.

* Created aws , oracle, azure classes to manage cloud config setting, argparser of cluster or other cloud specific functions.

* Added engine class to manage engine specific functions.

* Manage command line arguments accordingly w.r.t cloud. So for aws cloud only aws specific command line arguments will be shown. Similarly for azure and oracle.

bin/qds.py
```diff
@@ -524,6 +542,10 @@ def main():
                          default=False,
                          help="skip verification of server SSL certificate. Insecure: use with caution.")
 
+    optparser.add_option("--cloud", dest="cloud",
+                         default=os.getenv('CLOUD_PROVIDER'),
+                         help="cloud", choices=["AWS", "AZURE", "ORACLE_BMC"])
```
sourabh912: Where are we setting "aws" as default cloud provider?

qds_sdk/cloud/aws_cloud.py
```diff
@@ -0,0 +1,172 @@
...
+
+            master_elastic_ip: It is the Elastic IP address for attaching to the cluster master
+
+        '''
+
+        self.set_compute_config(use_account_compute_creds, compute_access_key, compute_secret_key, role_instance_profile)
+        self.set_location(aws_region, aws_availability_zone)
+        self.set_network_config(bastion_node_public_dns, persistent_security_groups, master_elastic_ip, vpc_id, subnet_id)
+
+
```
sourabh912: nit: Remove extra space.

qds_sdk/cloud/aws_cloud.py
```diff
@@ -0,0 +1,172 @@
...
+                           vpc_id=None,
+                           subnet_id=None):
+        self.network_config['bastion_node_public_dns'] = bastion_node_public_dns
+        self.network_config['persistent_security_groups'] = persistent_security_groups
+        self.network_config['master_elastic_ip'] = master_elastic_ip
+        self.network_config['vpc_id'] = vpc_id
+        self.network_config['subnet_id'] = subnet_id
+
+
+
```
sourabh912: nit: Please remove extra lines.

qds_sdk/cloud/aws_cloud.py
```diff
@@ -0,0 +1,172 @@
...
+                              persistent_security_groups=arguments.persistent_security_groups,
+                              bastion_node_public_dns=arguments.bastion_node_public_dns,
+                              master_elastic_ip=arguments.master_elastic_ip)
+
+    def create_parser(self, argparser):
+
+        # compute settings parser
+        compute_config = argparser.add_argument_group("compute config settings")
+        compute_creds = compute_config.add_mutually_exclusive_group()
+        compute_creds.add_argument("--enable_account_compute_creds",
```
sourabh912: Please replace underscore (`_`) with dash (`-`) in `enable_account_compute_creds `.

qds_sdk/cloud/aws_cloud.py
```diff
@@ -0,0 +1,172 @@
...
+                                   dest="use_account_compute_creds",
+                                   action="store_true",
+                                   default=None,
+                                   help="to use account compute credentials")
+        compute_creds.add_argument("--disable_account_compute_creds",
```
sourabh912: Same as above comment.

qds_sdk/cloud/azure_cloud.py
```diff
@@ -0,0 +1,208 @@
...
+            vnet_resource_group_name: vnet resource group name for azure
+
+            master_elastic_ip: It is the Elastic IP address for attaching to the cluster master
+
+        '''
+
+
+        self.set_compute_config(use_account_compute_creds, compute_tenant_id, compute_subscription_id, compute_client_id, compute_client_secret)
+        self.set_location(location)
+        self.set_network_config(bastion_node_public_dns, persistent_security_groups, master_elastic_ip, vnet_name, subnet_name, vnet_resource_group_name)
```
sourabh912: nit: Split the fn arguments in multiple lines (not necessarily one line for one argument).

bin/qds.py
tanishgupta1: When we define configuration inside qubole.py then we initialise aws as default cloud.

qds_sdk/cloud/aws_cloud.py
tanishgupta1: done

qds_sdk/cloud/aws_cloud.py
tanishgupta1: done

qds_sdk/cloud/aws_cloud.py
tanishgupta1: done

qds_sdk/cloud/azure_cloud.py
tanishgupta1: done

bin/qds.py
sourabh912: Specifying it inside qubole.py is fine. We should also set the default here as well just like we are setting defaults for `api_token`, `api_url` etc

qds_sdk/cloud/aws_cloud.py
```diff
@@ -0,0 +1,168 @@
...
+                                   dest="use_account_compute_creds",
+                                   action="store_true",
+                                   default=None,
+                                   help="to use account compute credentials")
+        compute_creds.add_argument("--disable-account-compute-creds",
+                                   dest="use_account_compute_creds",
+                                   action="store_false",
+                                   default=None,
+                                   help="to disable account compute credentials")
+
```
sourabh912: nit: Remove an extra line here. Also please take care of this thing elsewhere in the PR.

qds_sdk/clusterv2.py
```diff
@@ -0,0 +1,549 @@
...
+class ClusterCmdLine:
+
+    @staticmethod
+    def parsers(action):
+        argparser = argparse.ArgumentParser(
+            prog="qds.py cluster",
+            description="Cluster Operations for Qubole Data Service.")
+        subparsers = argparser.add_subparsers(title="Cluster operations")
+
+        create = subparsers.add_parser("create", help="Create a new cluster")
```
sourabh912: Shouldn't we add subparsers based on the action type? For example we should add `create` subparser only if `action == "create"`.

qds_sdk/engine.py
```diff
@@ -0,0 +1,196 @@
...
+                               spark_version=arguments.spark_version,
+                               custom_spark_config=arguments.custom_spark_config,
+                               dbtap_id=arguments.dbtap_id,
+                               fernet_key=arguments.fernet_key,
+                               overrides=arguments.overrides)
+
+    @staticmethod
+    def engine_parser(argparser):
+        engine_group = argparser.add_argument_group("engine settings")
+        engine_group.add_argument("--flavour",
```
sourabh912: Where are we using this engine `flavour`?

qds_sdk/clusterv2.py
```diff
@@ -0,0 +1,549 @@
...
+        cluster_request.update(util._make_minimal(cluster_info.__dict__))
+        return cluster_request
+
+class ClusterInfoV2(object):
+    """
+    qds_sdk.ClusterInfoV2 is the class which stores information about a cluster_info.
+    You can use objects of this class to create/update/clone a cluster.
+    """
+
+    def __init__(self, label):
```
sourabh912: Why are we passing only `label` in the constructor ? I believe we should pass other member variables as well and their default can be set to `{}`.

qds_sdk/clusterv2.py
```diff
@@ -0,0 +1,549 @@
...
+
+        # cloud config parser
+        cloud = Cloud.get_cloud()
+        cloud.create_parser(subparser)
+
+        # cluster info parser
+        ClusterInfoV2.cluster_info_parser(subparser, action)
+
+        # engine config parser
+        Engine().engine_parser(subparser)
```
sourabh912: Why do we need to create `Engine` object ? Can't we make `engine_paser`  a static or a class method ?

qds_sdk/clusterv2.py
```diff
@@ -0,0 +1,549 @@
...
+            ClusterCmdLine.create_update_clone_parser(clone, action="clone")
+            clone.set_defaults(func=ClusterV2.clone)
+
+        return argparser
+
+    @staticmethod
+    def create_update_clone_parser(subparser, action=None):
+
+        # cloud config parser
+        cloud = Cloud.get_cloud()
```
sourabh912: Would this call return new cloud object everytime ?  If yes, then I think it is highly inefficient and error prone.
You can check `qubole.py` for example. It has cached an agent and reusing it.

qds_sdk/util.py
```diff
@@ -142,3 +142,31 @@ def underscore(word):
...
+
+def _read_file(file_path):
+    file_content = None
+    if file_path is not None:
+        try:
+            with open(file_path) as f:
+                file_content = f.read()
+        except IOError as e:
+            sys.stderr.write("Unable to read %s: %s\n" % (file_path, str(e)))
+            sys.exit(1)
```
sourabh912: I think you should not exit here. Instead you should throw IOError here and let its caller handle the exception (if at all).

tests/test_clusterv2.py
```diff
@@ -0,0 +1,362 @@
...
+from mock import Mock
+import tempfile
+sys.path.append(os.path.join(os.path.dirname(__file__), '../bin'))
+import qds
+from qds_sdk.connection import Connection
+from test_base import print_command
+from test_base import QdsCliTestCase
+from qds_sdk.cloud.cloud import Cloud
+
+class TestClusterCreate(QdsCliTestCase):
```
sourabh912: Can you write few tests apart from cluster clone, create and update ?

tanishgupta1: Other than cluster clone , create and update there are no code changes . For other cluster apis we don't use version. Even if the the user pass v2 or v1.3 it will use v1.2 only.

qds_sdk/clusterv2.py
tanishgupta1: as discussed I have removed get cloud function and renamed get_cloud_config to get_cloud which will fetch cloud config parser of cloud instead of cloud name. It has no relation with qubole.py

qds_sdk/cloud/cloud.py
```diff
@@ -0,0 +1,16 @@
...
+        cloud = Qubole.cloud
+        if cloud == "azure":
+            import qds_sdk.cloud.azure_cloud
+            return qds_sdk.cloud.azure_cloud.AzureCloud()
+        elif cloud == "oracle_bmc":
+            import qds_sdk.cloud.oracle_bmc_cloud
+            return qds_sdk.cloud.oracle_bmc_cloud.OracleBmcCloud()
+        else:
+            import qds_sdk.cloud.aws_cloud
+            return qds_sdk.cloud.aws_cloud.AwsCloud()
```
sourabh912: We should define functions `set_cloud_config_from_arguments` and `create_parser ` in cloud.py so that any class which extends this class in future knows that it has to implement these functions.

qds_sdk/engine.py
tanishgupta1: You can check on the same file that we are using engine flavour to define the engine type of cluster as mentioned in new v2 api changes.

qds_sdk/clusterv2.py
tanishgupta1: For old api v1.2 and v1.3 we are doing like that only. It is because label is the primary and mandatory parameter for cluster apis call while creating , updating(we can use label instead of id) and cloning cluster. It will also help customers that definition is same and use it as same like we have in old cluster apis.

bin/qds.py
tanishgupta1: Yes I am doing that only. For api_token and api_url we use env variables as default which is same as above for cloud too :).

sourabh912: I meant this code at line Number 564: 

`     
        if options.api_url is None: 
          options.api_url = "https://api.qubole.com/api/"
`
We should add similar check for cloud.

tests/test_clusterv2.py
tanishgupta1: done

qds_sdk/cloud/cloud.py
tanishgupta1: done

qds_sdk/clusterv2.py
tanishgupta1: done

qds_sdk/clusterv2.py
tanishgupta1: done

bin/qds.py
tanishgupta1: done

qds_sdk/cloud/cloud.py
```diff
@@ -0,0 +1,22 @@
...
+
+    def create_parser(self):
```
sourabh912: Function signature is incorrect. In aws_cloud.py, azure_cloud.py etc the signature id `def create_parser(self, argparser)`.

qds_sdk/cloud/cloud.py
```diff
@@ -0,0 +1,22 @@
...
+        return NotImplemented
+
+    def set_cloud_config_from_arguments(self):
```
sourabh912: Same as above comment. Please change the signature fn.

qds_sdk/cloud/oracle_bmc_cloud.py
```diff
@@ -0,0 +1,204 @@
...
+                              compartment_id=arguments.compartment_id,
+                              image_id=arguments.image_id,
+                              vcn_id=arguments.vcn_id,
+                              storage_tenant_id=arguments.storage_tenant_id,
+                              storage_user_id=arguments.storage_user_id,
+                              storage_key_finger_print=arguments.storage_key_finger_print,
+                              storage_api_private_rsa_key=arguments.storage_api_private_rsa_key)
+
+    def create_parser(self, argparser):
+
```
sourabh912: nit: Remove this extra line.

qds_sdk/cloud/oracle_bmc_cloud.py
```diff
@@ -0,0 +1,204 @@
...
+        # location settings parser
+        location_group = argparser.add_argument_group("location config settings")
+        location_group.add_argument("--oracle-region",
+                                    dest="region",
+                                    help="region to create the cluster in", )
+        location_group.add_argument("--oracle-availability-zone",
+                                    dest="availability_domain",
+                                    help="availability zone to" +
+                                         " create the cluster in", )
+
```
sourabh912: nit: Remove extra line. Please take care of this at all places in this PR.

qds_sdk/clusterv2.py
```diff
@@ -0,0 +1,549 @@
...
+                    It implies that slave nodes can be of different instance types
+
+                `custom_tags` : Custom tags to be set on all instances
+                    of the cluster. Specified as JSON object (key-value pairs)
+
+                `datadog_api_token` : Specify the Datadog API token to use the Datadog monitoring service
+
+                `datadog_app_token` : Specify the Datadog APP token to use the Datadog monitoring service
+
+
```
sourabh912: Please remove these extra lines.

qds_sdk/qubole.py
```diff
@@ -56,6 +56,11 @@ def configure(cls, api_token,
         else:
             cls.poll_interval = poll_interval
         cls.skip_ssl_cert_check = skip_ssl_cert_check
+        if cloud is not None:
+            cls.cloud = cloud.lower()
```
sourabh912: Why would  cloud be `None` ? Isn't `aws` its default value ?

qds_sdk/cloud/oracle_bmc_cloud.py
tanishgupta1: This is done intentionally for separation of different parser for better visibility.

qds_sdk/clusterv2.py
```diff
@@ -0,0 +1,545 @@
...
+                                      idle_cluster_timeout=arguments.idle_cluster_timeout,
+                                      disk_count=arguments.count,
+                                      disk_type=arguments.disk_type,
+                                      disk_size=arguments.size,
+                                      upscaling_config=arguments.upscaling_config,
+                                      enable_encryption=arguments.encrypted_ephemerals,
+                                      customer_ssh_key=customer_ssh_key)
+
+        #  This will set cloud config settings
+        cloud_config = Cloud.get_cloud()
```
sourabh912: I am still not convinced with way we are getting cloud object. `Cloud.get_cloud()` will always return new object of a cloud. I think we can move `get_cloud()` in `qubole.py`. It can look like: 

``` python
    def get_cloud(cls, cloud_provider_name=None):
       if cloud_provider_name:
         // create a new cloud object here because cloud provider name is explicitly passed 
       else:
        // return cls.cloud it is already set. 
         if cls.cloud is not None: 
             return cls.cloud;
          // create it, set it at class level and return that
         if (cls.cloud_provider == "aws")
            cls.cloud = new AwsCloud()
         elif(cls.cloud_provider == "azure") 
            cls.cloud = new AzureCloud()
         return cls.cloud; 
```

Set cloud_provider_name from the options passed by the user. 
This approach is similar to how we cache connection agent

tanishgupta1: Updated the function. I have handled it quite differently. In qubole configuration we give cloud details too , then on the basis of this cloud we will initialise cls.cloud_config. Now this cloud_config can be used anywhere like we use cache_agent. Also if the user want to change cloud config then they can always call get_cloud method by passing cloud details explicitly.

qds_sdk/qubole.py
```diff
@@ -87,3 +91,18 @@ def agent(cls, version=None):
           cls.cached_agent = Connection(cls._auth, cls.rest_url, cls.skip_ssl_cert_check)
 
         return cls.cached_agent
+
+    @classmethod
+    def get_cloud(cls, cloud=None):
```
sourabh912: What if someone passed cloud = `"abc"` in this fn, we would return `cls.cloud_config` which is totally absurd.

qds_sdk/clusterv2.py
```diff
@@ -0,0 +1,545 @@
...
+            clone = subparsers.add_parser("clone", help="Clone a cluster from an existing one")
+            ClusterCmdLine.create_update_clone_parser(clone, action="clone")
+            clone.set_defaults(func=ClusterV2.clone)
+
+        return argparser
+
+    @staticmethod
+    def create_update_clone_parser(subparser, action=None):
+        # cloud config parser
+        cloud = Qubole.cloud_config
```
sourabh912: To get `cloud_config` you should not call `Qubole.cloud_config`. Instead you should call `Qubole.get_cloud()`. 

This way it would either return cached cloud config object or it would return a new object (if you pass cloud name in `get_cloud()` fn). Please see how Qubole.agent() is being used at various places in the code. I believe `Qubole.get_cloud()` should be used similarly.

tanishgupta1: Right now also it will return cached object. I will be calling get_cloud function once during configuration . And after that user can use Qubole.cloud_config everywhere accordingly. If the user want to change it then they can directly call get_cloud by passing cloud_provider.

sourabh912:
This PR is merged into `unreleased` branch. @tanishgupta1 : Thanks for bearing with me through out the PR :)

qds_sdk/cloud/aws_cloud.py
```diff
@@ -0,0 +1,166 @@
...
+                Do not specify this value for a public subnet.
+
+            master_elastic_ip: It is the Elastic IP address for attaching to the cluster master
+
+        '''
+
+        self.set_compute_config(use_account_compute_creds, compute_access_key,
+                                compute_secret_key, role_instance_profile)
+        self.set_location(aws_region, aws_availability_zone)
+        self.set_network_config(bastion_node_public_dns, persistent_security_groups,
```
anumahadevan: Why don't we have self.set_storage_config in aws. Have we taken care of duplicating the compute config as storage config in the aws v2 payload?

PR #190, opened 2017-04-29
dev: fix: SDK-191: Access denied errors in some situation while using boto2

harshshah87:
We get access  denied errors in some situations for boto2. Adding a catch to make sure we retry in such situations.

### DIFF
--- a/qds_sdk/commands.py
+++ b/qds_sdk/commands.py
@@ -436,6 +436,8 @@ class SparkCommand(Command):
 
     optparser.add_option("--sql", dest="sql", help="sql for Spark")
 
+    optparser.add_option("--note-id", dest="note_id", help="Id of the Notebook to run.")
+
     optparser.add_option("-f", "--script_location", dest="script_location",
                          help="Path where spark program to run is stored. Has to be a local file path")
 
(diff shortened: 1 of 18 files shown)


qds_sdk/commands.py
```diff
@@ -1316,7 +1316,18 @@ def _is_complete_data_available(bucket_paths, num_result_dir):
...
         log.info("Downloading file from %s" % s3_path)
         if delim is None:
-            key_instance.get_contents_to_file(fp)  # cb=_callback
+            try:
+                key_instance.get_contents_to_file(fp)  # cb=_callback
+            except boto.exception.S3ResponseError as e:
+                if (e.status == 403):
+                    # SDK-191, boto gives an error while fetching the objects using versions which happens by default
+                    # in the get_contents_to_file() api. So attempt one without specifying version.
+                    log.info("Access denied while fetching the s3 object. Retrying without specifying the version....")
```
sourabh912: nit: log.error ?

sourabh912: Also have we checked with AWS (or read in their sdk docs) of this behaviour ? This seems like a hack. Are we able to reproduce it in house ?

msumit:
lgtm

harshshah87:
Merged

PR #191, opened 2017-05-04
Adding changes for supporting remote script location

tarrygo:
These changes are for supporting S3 scripts for spark commands.

### DIFF
--- a/tests/test_command.py
+++ b/tests/test_command.py
@@ -428,10 +428,27 @@ class TestSparkCommand(QdsCliTestCase):
             qds.main()
 
     def test_submit_script_location_aws(self):
-        sys.argv = ['qds.py', 'sparkcmd', 'submit', '--script_location', 's3://bucket/path-to-script']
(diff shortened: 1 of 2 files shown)


tests/test_command.py
```diff
@@ -428,10 +428,27 @@ def test_submit_notebook_with_program(self):
             qds.main()
 
     def test_submit_script_location_aws(self):
```
sourabh912: @tarrygo : Please add a test case for unsupported file extension.

sourabh912: @tarrygo : Also please add tests for aws scala, R and sql just like we have tests for local script location.

tarrygo:
Added testcase to check for language and failure case.

sourabh912:
Merged into unreleased. Commit Link: https://github.com/qubole/qds-sdk-py/commit/cb29fd258faf29a851b7d1c0b4abd3a1f6076b92

PR #201, opened 2017-06-21
Sdk 218: Adding data_disk_count_and data_disk_size param for OPC

tarrygo:
Adding support for data_disk_count and data_disk_size in sdk. These params will taken from clusterV2 interface and will be added in storage_config. The cluster V2 create API expects these parameter in storage_config


abhijitjaiswal:
Merged to Unreleased branch.

PR #202, opened 2017-07-12
Fix using commands with StringIO

rowillia:
From the docs:

"This is not part of the TextIOBase API and may not exist in some implementations."

https://docs.python.org/3/library/io.html#io.TextIOBase.buffer


tomhettinger:
Would love to see this merged!  Would like to download straight to StringIO in memory, rather than writing/reading from disk (Python3).

harshshah87:
Approved. @adityaka , can we merge?

harshshah87:
Merged. c51e57c into unreleased branch.

PR #204, opened 2017-08-03
SDK-219 : DBExport command does not have option to pass Customercluster option

NamrataAShettar:
Added 'use_customer_cluster' and 'customer_cluster_label' options to DBExport and DBImport commands. Added the corresponding unit tests.

### DIFF
--- a/qds_sdk/commands.py
+++ b/qds_sdk/commands.py
@@ -977,6 +977,10 @@ class DbExportCommand(Command):
                          help="Modes 1 and 2: DbTap Id of the target database in Qubole")
     optparser.add_option("--db_table", dest="db_table",
                          help="Modes 1 and 2: Table to export to in the target database")
+    optparser.add_option("--use_customer_cluster", dest="use_customer_cluster", default=False,
+                         help="Modes 1 and 2: To use cluster to run command ")
+    optparser.add_option("--customer_cluster_label", dest="customer_cluster_label",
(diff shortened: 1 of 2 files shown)


msumit:
lgtm

qds_sdk/commands.py
```diff
@@ -977,6 +977,10 @@ class DbExportCommand(Command):
                          help="Modes 1 and 2: DbTap Id of the target database in Qubole")
     optparser.add_option("--db_table", dest="db_table",
                          help="Modes 1 and 2: Table to export to in the target database")
+    optparser.add_option("--use_customer_cluster", dest="use_customer_cluster", default=False,
```
sourabh912: @NamrataAShettar : Can you explain the api behaviour for the following:

1. use_customer_cluster is True but customer_cluster_label is not set. 
2. customer_cluster_label is set but use_customer_cluster is False

Should we perform these checks here instead of sending those requests to server?

Also, add tests covering these cases.

NamrataAShettar: The test cases have been added in next commit. Regarding the api behaviour:
1) It throws an error if use_customer_cluster is True but customer_cluster_label not set.
2) It accepts the request if use_customer_cluster is False but the customer_cluster_label is set but doesn't use the cluster. 
The following were discussed with @msumit and @sourabh912 .

msumit:
The feature already present in master. Closing this PR.

PR #209, opened 2017-11-07
Sdk 238 - Python SDK: delete command template failing with 404 Not Found

adityaka:
Added delete_template operation to the python sdk this adds a new call to the Templates Resource for deletion of templates. There is a unit test added along with it.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -47,3 +47,6 @@ tags
 
 # Idea IDE
 .idea
+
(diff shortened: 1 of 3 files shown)


harshshah87:
LGTM

adityaka:
Merged to the unreleased branch . Thanks for the review @harshshah87

PR #211, opened 2017-11-21
SDK-243: Issue while running two different command in different account using qds-sdk

adityaka:
Clear the cached agent if the configure is called with a different api token. 
This will enable sdk to use two different account tokens in same script


adityaka:
Merged to Unreleased branch closing the PR

PR #213, opened 2017-12-15
fix: user: SDK-250: Support for status api call using cluster v2 api version

anumahadevan:
Support for status api call using cluster v2 api version. Added a unit test also for the same.


tanishgupta1:
PR LGTM. I think this PR will handle other apis too apart from status ?

tanishgupta1:
As discussed remove the changes of account get creds from this PR. Raise another PR which will have changes of new get creds api with changes in commands.py too.

adityaka:
Merged to unreleased branch

PR #220, opened 2018-05-07
SDK-265: Adding a parameter per_page to the list() method

somani:
Tested it by using the sdk in my python scripts.


somani:
@adityaka Can you review please

msumit:
Looks like its being handled in https://github.com/qubole/qds-sdk-py/pull/257/files

PR #226, opened 2018-08-21
fix: dev: SDK-288: Bypass EC check logic in SDK

Joylal4896:
This change removes EC Check from SDK code.


qds_sdk/commands.py
```diff
@@ -1373,16 +1348,6 @@ def _is_complete_data_available(bucket_paths, num_result_dir):
         #It is a folder
         key_prefix = m.group(2)
         bucket_paths = bucket.list(key_prefix)
-        if not skip_data_avail_check:
```
msumit: need to remove all references of `skip_data_avail_check`, as there will be no use of this flag.

Joylal4896: sure, will do that

PR #230, opened 2018-09-07
added send kill in case of signal interruption

hidarapaneni:


chattarajoy:
The branch is out of date and has some additional commits from older branches, please rebase and open if necessary.

PR #235, opened 2018-09-14
fix: dev: SDK-275: Pass custom_ec2_tags in node_configuration for ClusterInfoV13

akaranjkar-qu:
`custom_ec2_tags` is nested under node_configuration instead of hadoop_settings for API v1.3. Updated ClusterInfoV13 to the same


msumit:
@akaranjkar-qu can you resolve the conflicts?

PR #239, opened 2018-10-18
SDK-307: Fix write_headers function

msumit:
- Fix while getting query schema from qlog
- Fix for headers which are being written twice

### DIFF
--- a/qds_sdk/commands.py
+++ b/qds_sdk/commands.py
@@ -262,11 +262,11 @@ class Command(Resource):
                 #fetch latest value of num_result_dir
                 num_result_dir = Command.find(self.id).num_result_dir
 
-                for s3_path in r['result_location']:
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -2,7 +2,7 @@
 import sys
 from setuptools import setup
 
-INSTALL_REQUIRES = ['requests >=1.0.3', 'boto >=2.45.0', 'six >=1.2.0', 'urllib3 >= 1.0.2', 'inflection >= 0.3.1']
+INSTALL_REQUIRES = ['requests >=1.0.3', 'boto >=2.45.0', 'six >=1.2.0', 'urllib3 >= 1.0.2, <1.23', 'inflection >= 0.3.1']
```
tanishgupta1: why we need to change this version ?

msumit: travis was failing.. will remove it..

PR #241, opened 2018-10-23
fix: SDK310 - remove unsupported SIGQUIT in Windows

aimboden:
signal.SIGQUIT throws an unsupported error under Windows. This PR adresses it.


aimboden:
Not sure why Travis is failing on 2.6, seems unrelated to this PR. Do we even need to support all the way back to 2.6? Seems that this is causing a lot of special casing in the code and tests.

PR #243, opened 2018-11-07
chg: dev: SDK-233: Add pool option for Hive, Hadoop, Spark, Shell and Pig commands

akaranjkar-qu:


msumit:
@akaranjkar-qu can you add a spec where `pool` parameter is being used?

msumit:
@akaranjkar-qu also 32 tests are failing cause of this.

PR #255, opened 2018-12-19
SDK-326: Displaying a proper error message in case of failed API calls 

Joylal4896:
Display a predefined error message when a non JSON decode-able response is received from API request.

### DIFF
--- a/qds_sdk/connection.py
+++ b/qds_sdk/connection.py
@@ -99,13 +99,15 @@ class Connection:
         else:
             raise NotImplemented
 
+        self._validate_json(r)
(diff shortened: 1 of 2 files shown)


qds_sdk/connection.py
```diff
@@ -99,13 +99,15 @@ def _api_call_raw(self, req_type, path, data=None, params=None):
         else:
             raise NotImplemented
 
+        self._validate_json(r)
         self._handle_error(r)
```
msumit: static methods won't need to be called via self right?

PR #257, opened 2019-01-18
chg: dev: SDK-330: Add page and per-page option to Cluster and ClusterV2

akaranjkar-qu:
- Added page and per_page option to Cluster
- Added page and per_page option to ClusterV2
- Fixed a bug with Cluster#list method where providing state param results in `TypeError: string indices must be integers`

### DIFF
--- a/qds_sdk/cluster.py
+++ b/qds_sdk/cluster.py
@@ -52,9 +52,10 @@ class Cluster(Resource):
         group.add_argument("--state", dest="state", action="store",
                            choices=['up', 'down', 'pending', 'terminating'],
                            help="list only clusters in the given state")
-        group.add_argument("--page", dest="page", action="store", type=int,
+        pagination_group = group.add_argument_group()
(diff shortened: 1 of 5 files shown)


qds_sdk/cluster.py
```diff
@@ -68,14 +72,26 @@ def list(cls, state=None):
             List of clusters satisfying the given criteria
         """
         conn = Qubole.agent()
+        params = {}
+        if page:
+            params['page'] = page
+        if per_page:
+            params['per_page'] = per_page
+        params = None if not params else params
+        cluster_list = conn.get(cls.rest_entity_path, params=params)
```
msumit: afaik clusters v1 does not support pagination

akaranjkar-qu: Yes that's right. When using v1.2, it just returns all the clusters. Should I modify it to log a warning that v1.2 doesn't support pagination and will list all the clusters?

akaranjkar-qu:
As discussed, logged a warning when pagination options are provided with API v1.2

PR #278, opened 2019-07-18
fix: dev: SDK-364: Oracle Multi AD changes required in qds-sdk

lingarajg:
Added support for availability_domain_info_map required for Oracle clusters.

### DIFF
--- a/qds_sdk/cloud/oracle_bmc_cloud.py
+++ b/qds_sdk/cloud/oracle_bmc_cloud.py
@@ -131,6 +131,10 @@ class OracleBmcCloud(Cloud):
         self.storage_config['block_volume_size'] = block_volume_size
 
     def set_cloud_config_from_arguments(self, arguments):
+        try:
+            arguments.availability_domain_info_map = ast.literal_eval(arguments.availability_domain_info_map)
+        except Exception as e:
+            raise Exception("Invalid List format for availability_domain_info_map: %s" % e.message)
         self.set_cloud_config(compute_tenant_id=arguments.compute_tenant_id,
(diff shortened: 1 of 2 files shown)


qds_sdk/cloud/oracle_bmc_cloud.py
```diff
@@ -133,6 +131,11 @@ def set_storage_config(self,
         self.storage_config['block_volume_size'] = block_volume_size
 
     def set_cloud_config_from_arguments(self, arguments):
+        if arguments.availability_domain_info_map:
+            try:
+                arguments.availability_domain_info_map = ast.literal_eval(arguments.availability_domain_info_map)
+            except Exception as e:
+                raise Exception("Invalid List format for availability_domain_info_map: %s" % e.message)
```
akaranjkar-qu: If a dict is passed for `availability_domain_info_map` it will not raise an exception. So we aren't correctly verifying that the string evaluates to a list only. Better to assert that `arguments.availability_domain_info_map` evaluates to a list and not some other type. Something like `assert isinstance(arguments.availability_domain_info_map, list)`.

Alternatively, we can change the availability_domain_info_map argument to accept more than one argument (with nargs='+') and then json.loads them to form a list of dicts.

```
        network_config_group.add_argument("--availability-domain-info-map",
                                          dest="availability_domain_info_map", nargs = '+'
                                          help="availability domain and subnet mapping for the cluster")
```

lingarajg: Had tried the alternate option initially but did not work, will try with the first option

PR #280, opened 2019-07-30
updating requests and urllib3

brickyard:
Security scans are showing CRLF injection vulnerabilities in the older versions of requests and urllib3.  

included unittests pass and manual spot testing of examples in the README.rst succeed.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,5 @@
 language: python
 python:
-  - "2.6"
   - "2.7"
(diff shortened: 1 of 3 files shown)


chattarajoy:
thanks for the PR @brickyard. I have added a few other changes to fix the Travis CI and also move requests to `2.21.0` as it is the latest version supported on `Python 3.4`. I hope that would also resolve the security vulnerability.

PR #288, opened 2019-10-30
spark streaming cluster support

saiyam1712:


chattarajoy:
@saiyam1712  please change the destination branch to unreleased and rebase your branch accordingly.

Changes LGTM but, are these changes enough to create a new cluster? please test your changes once and probably some basic operations on that cluster. Also would be good to add a test case.

PR #292, opened 2019-12-06
SDK-383: Add Retries to all Get requests in case of 500 Response

chattarajoy:
We often see cases when Airflow Jobs fail because the polling API returns a 500 Response. This change adds retries to all get calls when a 500 Response is returned as a result of heavy load on Qubole's Webapp.

### DIFF
--- a/qds_sdk/connection.py
+++ b/qds_sdk/connection.py
@@ -50,11 +50,11 @@ class Connection:
             self.session_with_retries = requests.Session()
             self.session_with_retries.mount('https://', MyAdapter(max_retries=3))
 
-    @retry((RetryWithDelay, requests.Timeout), tries=6, delay=30, backoff=2)
+    @retry((RetryWithDelay, requests.Timeout, ServerError), tries=6, delay=30, backoff=2)
     def get_raw(self, path, params=None):
         return self._api_call_raw("GET", path, params=params)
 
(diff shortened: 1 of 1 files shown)


qds_sdk/connection.py
```diff
@@ -50,11 +50,11 @@ def __init__(self, auth, rest_url, skip_ssl_cert_check, reuse=True):
             self.session_with_retries = requests.Session()
             self.session_with_retries.mount('https://', MyAdapter(max_retries=3))
 
-    @retry((RetryWithDelay, requests.Timeout), tries=6, delay=30, backoff=2)
+    @retry((RetryWithDelay, requests.Timeout, ServerError), tries=6, delay=30, backoff=2)
     def get_raw(self, path, params=None):
         return self._api_call_raw("GET", path, params=params)
 
-    @retry((RetryWithDelay, requests.Timeout), tries=6, delay=30, backoff=2)
+    @retry((RetryWithDelay, requests.Timeout, ServerError), tries=6, delay=30, backoff=2)
```
sunasra: This will apply to all network call? @chattarajoy

chattarajoy: only get requests. Retrying put/post/delete requests might case issues, hence added these to get requests only.

PR #294, opened 2020-01-17
new:dev:INFRA-2843:Add Retries and Delay Options in SDKs

shekharsaurabh:
As part of API throttling phase-2 dev, we are extending the capability to add retries and delay options for the user which can be consumed in case of RetryableExceptions


shekharsaurabh:
Raised - https://github.com/qubole/qds-sdk-py/pull/295

PR #295, opened 2020-01-17
new:dev:SDK-386:Add Retries and Delay Options in python-sdk

shekharsaurabh:
As part of API throttling phase-2 dev, we are extending the capability to add retries and delay options for the user which can be consumed in case of RetryableExceptions.


qds_sdk/qubole.py
```diff
@@ -21,6 +21,8 @@ class Qubole:
     """
 
     MIN_POLL_INTERVAL = 1
+    RETRIES_CAP = 10
```
yogesh2021: Try to come up with optimal value of retries and delay considering we want to finish all retries within 5 minutes.

qds_sdk/qubole.py
```diff
@@ -21,6 +21,8 @@ class Qubole:
     """
 
     MIN_POLL_INTERVAL = 1
+    RETRIES_CAP = 6
```
shekharsaurabh: Calculated the values as follow- 
With backoff factor as 2 and retry count of 6, we will be retrying 5 times, till the retry_count is >1, decrementing it on each retry. Keeping base_delay as 10sec,
Thus, all retries will be completed within
10 + 20 + 40 + 80 + 160 = 310sec (around 5 minutes)

qds_sdk/connection.py
```diff
@@ -165,6 +191,9 @@ def _handle_error(response):
         elif code == 449:
             sys.stderr.write(response.text + "\n")
             raise RetryWithDelay(response, "Data requested is unavailable. Retrying ...")
+        elif code == 429:
```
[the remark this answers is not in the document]
shekharsaurabh: Addressed

chattarajoy:
@shekharsaurabh , please merge the latest unreleased to your branch so that we can merge the changes.

PR #303, opened 2020-02-17
new:dev:SDK-387:Add hive version support 

satyabolnedi:


qds_sdk/engine.py
```diff
@@ -71,7 +75,10 @@ def set_engine_config(self,
 
         '''
 
-        self.set_hadoop_settings(custom_hadoop_config, use_qubole_placement_policy, is_ha, fairscheduler_config_xml, default_pool, enable_rubix)
+        self.set_hadoop_settings(custom_hadoop_config, use_qubole_placement_policy,
+                                 is_ha, fairscheduler_config_xml,
+                                 default_pool, enable_rubix)
+        self.set_hive_settings(hive_version)
```
tarrygo: Verify the value of hive_version column in hive_cluster_configs with and without this change.

satyabolnedi: image

@tarrygo above is the screenshot from DB with the updated sdk changes and hive version being updated with 3.1

PR #313, opened 2020-03-30
fix: dev: SDK-397: add dbtap config for mlflow cluster

santosh1994:
Screenshot 2020-03-30 at 3 35 05 PM

### DIFF
--- a/qds_sdk/engine.py
+++ b/qds_sdk/engine.py
@@ -33,7 +33,8 @@ class Engine:
                           airflow_python_version=None,
                           is_ha=None,
                           enable_rubix=None,
-                          mlflow_version=None):
+                          mlflow_version=None,
+                          mlflow_dbtap_id=None):
         '''
 
         Args:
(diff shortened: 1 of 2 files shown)


qds_sdk/engine.py
```diff
@@ -253,4 +263,8 @@ def engine_parser(argparser):
                                            dest="mlflow_version",
                                            default=None,
                                            help="mlflow version for mlflow cluster", )
+        mlflow_settings_group.add_argument("--mlflow-dbtap-id",
+                                           dest="dbtap_id",
```
chattarajoy: use a different var named `mlflow_dbtap_id `

qds_sdk/engine.py
```diff
@@ -78,8 +78,16 @@ def set_engine_config(self,
                                  default_pool, enable_rubix)
         self.set_presto_settings(presto_version, custom_presto_config)
         self.set_spark_settings(spark_version, custom_spark_config)
-        self.set_airflow_settings(dbtap_id, fernet_key, overrides, airflow_version, airflow_python_version)
-        self.set_mlflow_settings(mlflow_version)
+        # because of same config dbtap_id used in mlflow and airflow,
+        # configs are getting created for both cluster type
+        if mlflow_version:
```
chattarajoy: can't we use a different var named `mlflow_dbtap_id ` for mlflow_version?

qds_sdk/engine.py
santosh1994: done

qds_sdk/engine.py
santosh1994: done

PR #321, opened 2020-06-01
SDK-407, JUPY-884: Add '--upload-to-source' param for JupyterNotebookCommand

tarungavara:
Airflow does not support boolean flags. Thus, we are removing '--skip-upload-to-source' option in JupyterNotebookCommand and adding '--upload-to-source' option.

The valid arguments to this option include 'True' and 'False'. Passing any other value would raise a ParseError.

### DIFF
--- a/qds_sdk/commands.py
+++ b/qds_sdk/commands.py
@@ -1387,11 +1387,10 @@ class JupyterNotebookCommand(Command):
     optparser.add_option("--print-logs-live", action="store_true",
                          dest="print_logs_live", default=False, help="Fetch logs \
                          and print them to stderr while command is running.")
-    optparser.add_option("--skip-upload-to-source", action="store_false",
-                         dest="upload_to_source", default=True, help="Do not \
-                         upload notebook to source after completion of execution")
-    optparser.add_option("--upload-to-source", dest="upload_to_source", help="Do not \
-                         upload notebook to source after completion of execution")
+    optparser.add_option("--upload-to-source", dest="upload_to_source", default='True',
+                         help="Upload notebook to source after completion of \
+                         execution. Specify value as either True or False. Default \
+                         value is True.")
 
     @classmethod
     def parse(cls, args):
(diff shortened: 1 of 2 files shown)


qds_sdk/commands.py
```diff
@@ -1417,6 +1418,16 @@ def parse(cls, args):
...
                 options.retry = int(options.retry)
+            if options.upload_to_source is not None:
+                options.upload_to_source = options.upload_to_source.lower()
+                if options.upload_to_source == 'true':
+                    options.upload_to_source = True
+                elif options.upload_to_source == 'false':
+                    options.upload_to_source = False
+                else:
+                    msg = 'Upload to Source parameter takes value of either True \
+                    or False only.'
```
chattarajoy, line 1429:
```suggestion
                    msg = "Upload to Source parameter takes a value of either 'true' \
                    or 'false' only."
```

qds_sdk/commands.py
```diff
@@ -1387,9 +1387,10 @@ class JupyterNotebookCommand(Command):
     optparser.add_option("--print-logs-live", action="store_true",
                          dest="print_logs_live", default=False, help="Fetch logs \
                          and print them to stderr while command is running.")
-    optparser.add_option("--skip-upload-to-source", action="store_false",
-                         dest="upload_to_source", default=True, help="Do not \
-                         upload notebook to source after completion of execution")
+    optparser.add_option("--upload-to-source", dest="upload_to_source", default='True',
+                         help="Upload notebook to source after completion of \
+                         execution. Specify value as either True or False. Default \
+                         value is True.")
```
chattarajoy, line 1393:
```suggestion
                         execution. Specify the value as either'true' or 'false'. Default \
                         value is 'true' ")
```

qds_sdk/commands.py
chattarajoy: otherwise, the message gives the impression that the user needs to pass `True` in python-format

PR #323, opened 2020-06-13
Ability to set azure cluster composition

harshshah87:


bin/qds.py
```diff
@@ -587,7 +587,7 @@ def main():
                          help="skip verification of server SSL certificate. Insecure: use with caution.")
 
     optparser.add_option("--cloud_name", dest="cloud_name",
-                         default=os.getenv('CLOUD_PROVIDER'),
+                         default=os.getenv('CLOUD_PROVIDER') or "AWS",
```
chattarajoy, line 590: better use getenv defaults
```suggestion
                         default=os.getenv('CLOUD_PROVIDER', "AWS"),
```

qds_sdk/cloud/cloud.py
```diff
@@ -5,3 +5,9 @@ def create_parser(self, argparser):
 
     def set_cloud_config_from_arguments(self, arguments):
         return NotImplemented
+
+    def set_composition_arguments(self, comp_group):
+        pass
```
chattarajoy, line 10: should we

```suggestion
        return NotImplemented
```
?

bin/qds.py
harshshah87: If env level default is not present, then only it will move to hard coded value. Else the code simply fails as no cloud is defined.

qds_sdk/cloud/cloud.py
harshshah87: If we go with that approach, we have to check for cloud when we set the composition. 
Rather than having an if condition to check whether the cloud supports, each cloud can set empty data or implement the function will sets the value. 
I implemented it this way to avoid that if conditions. In future, user has to implement the function and no need to edit the condition to allow cloud to set compositions.

bin/qds.py
harshshah87: my bad, i read the comment wrong. Will change it.

chattarajoy:
LGTM overall, @harshshah87 please let me know if this is ready to merge

qds_sdk/qubole.py
```diff
@@ -30,7 +30,7 @@ class Qubole:
     version = None
     poll_interval = None
     skip_ssl_cert_check = None
-    cloud_name = None
+    cloud_name = "aws"
```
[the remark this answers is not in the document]
harshshah87: Done

harshshah87:
@chattarajoy : We can merge now.

PR #325, opened 2020-06-19
Quest-674: Changes to support BYOC/BYOJ pipeline creation

siddharth1001:
Support BYOC/BYOJ pipeline creation in one go with notification channels.

Adding support for REST API.
Adding test for this case via command line will need some changes in the design of Pipelines support in sdk hence not doing the same in this PR.


qds_sdk/pipelines.py
```diff
@@ -339,23 +339,70 @@ def create(cls, pipeline_name, create_type, **kwargs):
...
-                {"name": pipeline_name, "status": "DRAFT",
-                 "create_type": create_type},
-            "type": "pipelines"}
-        }
-        url = Pipelines.rest_entity_path + "?mode=wizard"
+        url = Pipelines.rest_entity_path
+        if create_type is None:
+            raise ParseError("Provide create_type for Pipeline.", None)
+        if not kwargs or create_type == 1:
+            data = {
```
sunasra, line 353: are we handling for assist?

qds_sdk/pipelines.py
```diff
@@ -339,23 +339,70 @@ def create(cls, pipeline_name, create_type, **kwargs):
...
-            "attributes":
-                {"name": pipeline_name, "status": "DRAFT",
-                 "create_type": create_type},
-            "type": "pipelines"}
-        }
-        url = Pipelines.rest_entity_path + "?mode=wizard"
+        url = Pipelines.rest_entity_path
+        if create_type is None:
+            raise ParseError("Provide create_type for Pipeline.", None)
+        if not kwargs or create_type == 1:
```
sunasra, line 352: Are we handling for assisted?

siddharth1001: Yes

qds_sdk/pipelines.py
```diff
@@ -339,23 +339,70 @@ def create(cls, pipeline_name, create_type, **kwargs):
 
         Args:
             pipeline_name: Name to be given.
-            create_type: 1->Assisted, 2->Code, 3->Jar
+            create_type: 1->Assisted, 2->Jar, 3->Code
```
chattarajoy, line 342: hmm, breaking change, need to remember while releasing

qds_sdk/pipelines.py
```diff
@@ -339,23 +339,70 @@ def create(cls, pipeline_name, create_type, **kwargs):
...
+                            "status": "DRAFT",
+                            "create_type": create_type
+                        },
+                        "type": "pipeline"
+                    }
+            }
+            url = url + "?mode=wizard"
+        else:
+            data = {
+                "data": {
```
chattarajoy, line 366: can we store them in some constants? just makes it more readable here

qds_sdk/pipelines.py
```diff
@@ -339,23 +339,70 @@ def create(cls, pipeline_name, create_type, **kwargs):
...
+                                    "notification_channels": kwargs['channel_ids']
+                                }
+                            }
+                        }
+                    }
+                }
+            }
+            if create_type == 2:
+                data['data']['attributes']['properties']['jar_path'] = \
+                    kwargs['jar_path']
```
chattarajoy, line 393: try `kwargs.get` and handle with proper error in case the keys are not present, otherwise, we might. see key errors that the user might not understand

siddharth1001: We can use `kwargs.get` but I don't think we need to handle the case if keys are present or not over here. That should be done at the QDS API level.

qds_sdk/pipelines.py
siddharth1001: I didn't get you. Can you please explain with an example ?

qds_sdk/pipelines.py
siddharth1001: This was wrongly specified. I have corrected this.
create_type was always 2 for Jar and 3 for Code.

chattarajoy: okay got it

qds_sdk/pipelines.py
chattarajoy: yeah error from api also works

qds_sdk/pipelines.py
chattarajoy: the JSON part can be sent to some constant variables, otherwise, it makes a bit more complicated here to read.

chattarajoy: oh didn't see kwargs are being used in these blocks, it's fine. Let's just add `kawargs.get` should be fine

PR #331, opened 2020-10-12
SDK:419 : Adding support for passing extra arguments for fetching results

abhimarshal1:
This will enable the user to pass extra paramters as boto_args in command's result fetch which is then being passed to boto's connec_s3 function to give a flexibility to tune it accordingly.

### DIFF
--- a/qds_sdk/commands.py
+++ b/qds_sdk/commands.py
@@ -252,7 +252,7 @@ class Command(Resource):
         return r.text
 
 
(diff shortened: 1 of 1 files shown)


chattarajoy:
Can you please test this once with a bucket name which has dot in it and also try with a normal bucket as well.

chattarajoy:
Also, pls change the destination branch to unreleased and rebase your pvt branch with unreleased.

abhimarshal1:
> Also, pls change the destination branch to unreleased and rebase your pvt branch with unreleased.

Will be raising a new one over unreleased

PR #332, opened 2020-10-12
SDK:419 : Adding support for passing extra arguments for boto's conne…

abhimarshal1:
This will enable the user to pass extra paramters as boto_args in command's result fetch which is then being passed to boto's connec_s3 function to give a flexibility to tune it accordingly.


abhimarshal1:
@chattarajoy Will test this as you suggested.

Issue #119, opened 2015-08-26
Timeout is hard coded in connection wrapper


kaneda:
Timeout for net calls is hard coded to 300 seconds in https://github.com/qubole/qds-sdk-py/blob/master/qds_sdk/connection.py . This should be an option that is passed as part of params (and so on through https://github.com/qubole/qds-sdk-py/blob/master/qds_sdk/commands.py ) to allow a client to execute requests with variable timeout.

PR #336, opened 2021-05-03
Add timeout option in connection wrapper (#119)

choeh:
I have faced the same issue that was mentioned in #119  .

The default `300` seconds as timeout for connections to QDS was not big enough to run commands for heavy load SQL queries against QDS.

This PR enables users to specify their custom `timeout` value as a parameter and removes the hardcoded default value from the SDK. It furthermore helps overcome the `HTTPSConnectionPool(host='qubole.com', port=443): Read timed out. (read timeout=300)` error raised by `requests.exceptions.Timeout` when commands do not finish before the default `300s`.

### DIFF
--- a/bin/qds.py
+++ b/bin/qds.py
@@ -602,6 +602,11 @@ def main():
                          help="Number of re-attempts for an api-call in case of "
                               " retryable exceptions. Defaults to 7.")
 
+    optparser.add_option("--timeout", dest="timeout",
+                         type=int,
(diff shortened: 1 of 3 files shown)


choeh:
I just saw last commit is months ago... Do you think this can be merged in the shortterm and not in months @SandeepBelagavi, @chattarajoy, @akaranjkar-qu ? I highly appreciate a short response from anybody on this.

PR #340, opened 2023-10-14
fix comparisons to literals

### LINKED ISSUE #298: SyntaxWarning and deprecation warning over invalid escape sequences
```
./qds_sdk/role.py:23: DeprecationWarning: invalid escape sequence \]
  help="Policy Statement example '[{\"access\":\"deny\", \"resource\": \"all\", \"action\": \"[\"create\",\"update\",\"delete\"\]\"}]'")
./qds_sdk/role.py:42: DeprecationWarning: invalid escape sequence \]
  help="Policy Statement example '[{\"access\":\"deny\", \"resource\": \"all\", \"action\": \"[\"create\",\"update\",\"delete\"\]\"}]'")
./qds_sdk/commands.py:1124: SyntaxWarning: "is" with a literal. Did you mean "=="?
  if options.mode is "1":
./qds_sdk/commands.py:1137: SyntaxWarning: "is" with a literal. Did you mean "=="?
  if options.db_update_mode is "updateonly":
./qds_sdk/commands.py:1424: SyntaxWarning: "is" with a literal. Did you mean "=="?
  if (total is 0) or (downloaded == total):
./tests/test_command.py:1950: DeprecationWarning: invalid escape sequence \$
  sys.argv = ['qds.py', 'dbtapquerycmd', 'submit', '--query', "select * from table_1 limit  \$limit\$",
./tests/test_command.py:1958: DeprecationWarning: invalid escape sequence \$
  'query': "select * from table_1 limit  \$limit\$",
./tests/test_command.py:1966: DeprecationWarning: invalid escape sequence \$
  sys.argv = ['qds.py', 'dbtapquerycmd', 'submit', '--query', "select * from table_1 limit  \$limit\$",
./tests/test_command.py:1974: DeprecationWarning: invalid escape sequence \$
  'query': "select * from table_1 limit  \$limit\$",
```

guidopetri:
Change comparisons to literals from `is` to `==`.

Comparisons to literals with `is` aren't Pythonic because `is` uses the underlying object ID to determine whether objects match, whereas `==` does a value comparison. By switching the comparisons to `==`, we have a more consistent set of comparisons, because we don't rely on object IDs but rather the values themselves, which is what we're looking for, anyhow.

This partially fixes #298 , clearing some of the `SyntaxWarning`s raised.

### DIFF
--- a/qds_sdk/commands.py
+++ b/qds_sdk/commands.py
@@ -1121,7 +1121,7 @@ class DbExportCommand(Command):
                 raise ParseError("dbtap_id and db_table are required",
                                  cls.optparser.format_help())
 
-            if options.mode is "1":
+            if options.mode == "1":
                 if options.hive_table is None:
                     raise ParseError("hive_table is required for mode 1",
                                      cls.optparser.format_help())
@@ -1134,7 +1134,7 @@ class DbExportCommand(Command):
                     raise ParseError("db_update_mode should either be left blank for append "
                                      "mode or be 'updateonly' or 'allowinsert'",
                                      cls.optparser.format_help())
-                if options.db_update_mode is "updateonly":
+                if options.db_update_mode == "updateonly":
                     if options.db_update_keys is None:
                         raise ParseError("db_update_keys is required when db_update_mode "
                                          "is 'updateonly'",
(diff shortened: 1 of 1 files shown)


guidopetri:
The blocking check seems to check the entire `commands.py` file, I'm not sure what to do about that - I don't really want to make a ton of edits to the 1500 lines in that file :(

